PR #1, opened 2015-06-11
small corrections to MakeFile

Mingling94:
the rm default_app should be a mv to rename it to app... I think.
@DavidVorick comment on this, it deals with your most recent commit 16 days ago but I had to do it to run test-ui. Haven't touched MakeFile manually before in my life yo

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,10 +5,10 @@ clean:
 
 dependencies:
 	npm install electron-prebuilt
-	mv node_modules/electron-prebuilt/dist/resources/default_app node_modules/electron-prebuilt/dist/resources/app
+	rm -rf node_modules/electron-prebuilt/dist/resources/default_app
 
 run:
-	rm -r node_modules/electron-prebuilt/dist/resources/app/
+	rm -rf node_modules/electron-prebuilt/dist/resources/app/
 	cp -R site/ node_modules/electron-prebuilt/dist/resources/app/
 	./node_modules/electron-prebuilt/dist/electron
 
--- a/Makefile
+++ b/Makefile
@@ -5,11 +5,11 @@ clean:
 
 dependencies:
 	npm install electron-prebuilt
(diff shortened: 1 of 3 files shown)


DavidVorick:
no it's supposed to remove it.

Makefile
```diff
@@ -5,7 +5,7 @@ clean:
 
 dependencies:
 	npm install electron-prebuilt
-	rm node_modules/electron-prebuilt/dist/resources/default_app
```
Mingling94, line 8: @DavidVorick, So at the very least, this needs a -rf flag.

Then a subsequent `make` yields: 
selection_007

DavidVorick:
does it need the 'f'? I guess I did forget the 'r'.

You could run `@rm -r ...`, which supresses the output, but I think `rm -rf` is the better choice.

Thanks for catching that.

site/main.js
```diff
@@ -17,10 +18,15 @@ app.on('ready', function() {
...
 		"height": 720,
 		"width": 1200,
+        "min-width": 800,
+        "min-height": 600,
 		"title": "Sia"
+
 	});
 	mainWindow.loadUrl('file://' + __dirname + '/html/index.html');
 
+	mainWindow.openDevTools();
```
DavidVorick: don't have this enabled in master.

site/main.js
```diff
@@ -17,10 +18,15 @@ app.on('ready', function() {
 	mainWindow = new BrowserWindow({
 		"height": 720,
 		"width": 1200,
+        "min-width": 800,
+        "min-height": 600,
```
DavidVorick: tabbing

Mingling94: Heh, on my atom editor these appeared to be in line

DavidVorick:
We should get a linter of some sort for javascript, something that works similarly to go fmt.

PR #2, opened 2015-06-12
alternate makefile corrections

Mingling94:
Again, dunno what I'm doing with this manual makefile editing shenanigans. I'm still reading into the structure of electron applications and stuff, but this is more congruent with your comments on the other PR

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,10 +5,10 @@ clean:
 
 dependencies:
 	npm install electron-prebuilt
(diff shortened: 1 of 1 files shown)


DavidVorick:
These corrections look fine, except you've got conflicting corrections in your other PR. Please try to keep related things in a single PR. Especially try not to have two conflicting PRs open at the same time, it's a huge headache to sort out sometimes.

PR #4, opened 2015-06-17
Restructuring

Mingling94:
Lots of changes. Majorly they are:
1. Makefile changes, look these over carefully and test it out.
   a. `make` now does dependencies and run.
   b. `make dist` should also produce working executables.
   c. does not copy code into node_modules/electron-prebuilt/... Instead just runs electron upon ./app/
2. package.json
   a. Edited for more detail. 
   b. The fact that it's actually used by the Makefile now, allowing for devDependencies to matter.
   c. npm commands with the ./app/ folder work.
3. READMEs explain all of the above


DEVELOPERS.md
```diff
@@ -1,4 +1,17 @@
-I am only certain about development flow for linux. If you are developing on another operating system, please expand this documentation.
+The top level folder is pertinent mostly for its Makefile which holds the bash
+logic to make distributables for each OS through a `make dist` command
 
-You will need npm and gconf. Most linux installations will already have gconf. You should also have tar and zip.
-To get started, run 'make dependencies && make'.
+The ./app/ folder holds all the code relevant to make test-ui into an npm module
```
DavidVorick: ./app/ makes it seem like the directory is executable.

It's sufficient to just say app/

DEVELOPERS.md
```diff
@@ -1,4 +1,17 @@
-I am only certain about development flow for linux. If you are developing on another operating system, please expand this documentation.
+The top level folder is pertinent mostly for its Makefile which holds the bash
+logic to make distributables for each OS through a `make dist` command
```
DavidVorick: missing a period

Makefile
```diff
@@ -1,37 +1,47 @@
...
+electron-download-page = https://github.com/atom/electron/releases/download
+electron-version = v0.28.1
 sia-version = v0.4.0
 
-all: run
+# runs the app
+all: dependencies run
 
+# cleans distributable files
 clean:
```
DavidVorick: if 'make dependencies' pulls in any folders, then 'make clean' should purge those as well.

Makefile
```diff
@@ -1,37 +1,47 @@
-# Version control
-electron-version = v0.28.0
+# Version control variables
+electron-download-page = https://github.com/atom/electron/releases/download
```
DavidVorick, line 2: this variable might be overkill, but I guess it's okay.

Makefile
DavidVorick: perhaps this:

``` make
clean: clean-node-modules
    rm -rf electron-* sia-v* Sia-*
```

You want `make clean` to remove old electron versions as well, not just the most recent.

Makefile
```diff
@@ -1,37 +1,47 @@
...
 
+# cleans distributable files
 clean:
 	rm -rf electron-$(electron-version)* sia-v* Sia-*
 
+# install node_modules
 dependencies:
-	npm install electron-prebuilt
-	rm -rf node_modules/electron-prebuilt/dist/resources/default_app
+	( cd app ; npm install )
```
DavidVorick: this feels like the wrong approach.

Makefile
```diff
@@ -1,37 +1,47 @@
...
 
+# before running, node_modules needs to be installed
```
DavidVorick: this comment should be implicit. It's expected that you run `make dependencies` before you run `make` if you haven't yet fetched the dependencies.

Makefile
```diff
@@ -42,25 +52,23 @@ dist-mac:
...
 	unzip -o electron-$(electron-version)-win32-x64.zip -d Sia-Windows-x64
 	mv Sia-Windows-x64/electron.exe Sia-Windows-x64/Sia.exe
-	cp -R site/ Sia-Windows-x64/resources/app/
+	cp -R app/ Sia-Windows-x64/resources/app/
 	zip sia-$(sia-version)-windows-x64.zip Sia-Windows-x64/*
 
-dist: dist-linux-32 dist-linux-64 dist-mac dist-windows-32 dist-windows-64
-
-.PHONY: all clean dependencies run dist-linux-32 dist-linux-64 dist-mac dist-windows-32 dist-windows-64 dist
+.PHONY: all clean dependencies run clean-node-modules dist dist-linux-32 dist-linux-64 dist-mac dist-windows-32 dist-windows-64
```
DavidVorick: put 'clean' and 'clean-node-modules' next to each other, also do this for their declarations earlier in the makefile.

I think that 'clean-node-modules' is not the best name. It's the low-level thing that you are doing, but what's the high level thing? Also, why do you only run it when going `make dist`? I'm not sure of its place in the project.

README.md
```diff
@@ -0,0 +1,27 @@
+# Test UI
+
+This is the user interface for [Sia](https://github.com/NebulousLabs/Sia), it is a desktop application based off the
+[electron (formerly named atom-shell)](https://github.com/atom/electron) framework.
```
DavidVorick: at this point we're far enough from the name change to just call it electron.

app/html/plugins.html
```diff
@@ -0,0 +1,7 @@
+<template id="plugin-template">
+  <li class="plugin">
+    <div class="author"></div>
+    <div class="logo"></div>
+    <div class="view"></div>
+  </li>
+</template>
```
DavidVorick, line 7: what is the purpose of this file?

app/js/main.js
```diff
@@ -7,7 +7,7 @@ var mainWindow = null;
 
 // Respond to all windows being closed.
 app.on('window-all-closed', function() {
-	if (process.platform != 'darwin') {
+	if (process.platform !== 'darwin') {
```
DavidVorick: what is the difference?

app/js/main.js
```diff
@@ -23,7 +23,7 @@ app.on('ready', function() {
 		"title": "Sia"
 
 	});
-	mainWindow.loadUrl('file://' + __dirname + '/html/index.html');
+	mainWindow.loadUrl('file://' + __dirname + '/../html/index.html');
```
DavidVorick: main.js is an exception and doesn't need to be in the 'js' folder. It's cleaner to have it out because that's where people will expect to see it.

app/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "Sia-UI",
+  "version": "0.4.0",
+  "main": "./js/main.js",
+  "author": "NebulousLabs",
+  "description": "A UI application for interfacing with Sia",
+  "license": "MIT",
+  "devDependencies": {
+    "electron-prebuilt": "latest",
```
DavidVorick: electron-prebuilt is not a dev dependency. It's a dependency for everyone.

app/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "jshint": "latest",
+    "mocha": "latest"
```
DavidVorick, line 10: mocha is likely also a dependency for everyone. It's pretty typical to run a battery of tests on the users machine when they are building from source.

app/js/viewLoader.js
```diff
@@ -0,0 +1,31 @@
+// viewLoader manages which view is currently being displayed in the mainbar of
+// the UI.
+var window = null;
+
+window.onload = function() {
+	var view = window.$('#view'),
+		sidebarOverviewButton = window.document.getElementById('sidebar-overview-button'),
+		sidebarLibraryButton = window.document.getElementById('sidebar-library-button'),
+		sidebarWalletButton = window.document.getElementById('sidebar-wallet-button');
```
DavidVorick, line 9: tabbing

Does the linter catch this?

We use tabs, not spaces. Is there a linter setting for that?

DavidVorick:
'make' and 'make dependencies' should be separate, I'm not sure why you wanted to combine them.

Makefile
Mingling94: `npm install` or the setup to call it from within `app`?

app/html/plugins.html
Mingling94, line 7: Messing around to try and see how I'd template up markup for reuse with different data. Guess I should take it out and into another branch?

app/js/main.js
Mingling94: Mostly identical except === and !== are faster than == and != because no type conversion is done. The former equalities are more strict and I think they should always be adopted. Also linting standards suggest this.

[See this stackoverflow answer](http://stackoverflow.com/a/359509/1026909)

PR #12, opened 2015-07-05
Reorganization

Mingling94:
Daemon.js isn't working in this but that's not important, the latest commit of reorganization/renaming is what I wanted to run past you, @DavidVorick

PR #13, opened 2015-07-06
Incorporate siad

Mingling94:
Same PR as #12 but with 2 more commits.


DavidVorick:
This PR does not appear to make any changes to the makefile. `make dependencies` should probably include cloning and building the Sia repo.

Makefile
```diff
@@ -1,43 +1,54 @@
...
+sia-download-page = https://github.com/NebulousLabs/Sia/releases/download/${sia-version}
+project = "Sia-UI" ${sia-ui-version}
+ui-path := $(shell pwd)
+sia-ui-dependencies = $(ui-path)/app/dependencies/Sia
 
 # runs the app
 all: run
 
-run:
+run: ;@echo "\n${project}: Running.....\n";
```
DavidVorick: `make run` doesn't need an echo string. The other ones need an echo string because they are called as a chain and each part of the chain can take a long time.

Makefile
```diff
@@ -1,43 +1,54 @@
...
 	npm start
 
+# simulates a fresh execution
 run-fresh: clean dependencies run
 
 # cleans distributable files & dependencies
-clean:
-	rm -rf electron-* sia-v* Sia-* node_modules
+clean: ;@echo "\n${project}: Cleaning modules, siad, and distributables.....\n";
```
DavidVorick: also does not need an echo string

Makefile
```diff
@@ -1,43 +1,54 @@
...
+	rm -rf electron-* sia-v* Sia-* node_modules app/dependencies/Sia
 
-# install node_modules
-dependencies:
+# install tools needed to run
+dependencies: ;@echo "\n${project}: Installing modules and siad.....\n";
```
DavidVorick: also does not need an echo string

Makefile
```diff
@@ -1,43 +1,54 @@
...
+	go get -u github.com/NebulousLabs/Sia/...
+	(cd $(GOPATH)/src/github.com/NebulousLabs/Sia && \
+	git fetch && git checkout $(sia-version) && \
+	make dependencies && make && \
+	test -d $(sia-ui-dependencies) || mkdir -p $(sia-ui-dependencies) && cp $(GOPATH)/bin/siad $(sia-ui-dependencies))
```
DavidVorick, line 30: can you explain this line to me?

Makefile
```diff
@@ -1,43 +1,54 @@
...
 	npm install
 
 # make distributables for each operating system, folders and executable files
 dist: dist-linux-32 dist-linux-64 dist-mac dist-windows-32 dist-windows-64
 
-dist-linux-32:
-	wget -nc $(electron-download-page)/$(electron-version)/electron-$(electron-version)-linux-ia32.zip
+dist-linux-32: ;@echo "\n${project}: Building Linux 32 bit.....\n";
+	wget -nc $(electron-download-page)-linux-ia32.zip
```
DavidVorick, line 37: @lukechampine what do you think of this way of doing echoing?

Makefile
Mingling94, line 30: Ya sure. If this is the first time calling `make dependencies`, the makefile will make sure the `sia-ui-dependencies` folder exists OR create it if it doesn't, then go on to copying the binary from the Gopath to there

DavidVorick:
Luke and I discussed, and we think that all the echo's should be removed from the makefile.

PR #16, opened 2015-07-13
Api ipc

Mingling94:

PR #17, opened 2015-07-15
Overview

Mingling94:
On top of #16, It's now pretty. As always, more redesigning


app/index.js
```diff
@@ -42,7 +41,7 @@ function startMainWindow() {
 	mainWindow.focus();
 
 	// DEVTOOL: Open the devtools.
-	//mainWindow.openDevTools();
+	// mainWindow.openDevTools();
```
agundy: Commented out function call?

app/js/UI.js
```diff
@@ -1,26 +1,52 @@
...
+	function updateClient(version) {
+		//Shell.openExternal('https://www.github.com/NebulousLabs/Sia-UI/releases');
+	}
+	*/
+
 	// adjustZoom() makes the app more readable on high dpi screens. 
 	// TODO: Take better approach, resolution doesn't mean high dpi. Though
 	// supposedly there's not a sure-fire way to find dpi on all platforms.
-	function adjustHighResZoom() {
+	function adjustHighResZoom(config) {
```
agundy, line 49: Perhaps instead of checking and adjusting automatically make it a setting? Not sure if this is already done or not.

app/js/daemonManager.js
```diff
@@ -3,53 +3,137 @@
...
-		API.setConfig(config);
+		address = config.siadAddress;
 		callback();
 	}
 
-	// start() starts the daemon as a long running background process. siad already
-	// bounces if the port is in use.
+	// start() starts the daemon as a long running background process
 	function start() {
+		console.log('starting siad');
```
agundy, line 69: Console logging works for now but should use some form of logging to file. When launching a regular executable no one will see the console logs

app/index.js
DavidVorick: We keep it there for development. It more or less gets uncommented any time you start playing around with the program, so it made sense to just keep it in master, commented out.

app/js/UI.js
DavidVorick, line 49: I'm not 100% sure of the order, but I believe:
1. check if config exists. if so, use that and be done
2. config doesn't exist, figure out the reasonable default for the zoom

you can adjust the zoom by changing the config file.

app/index.js
Mingling94: Actually, I'm going to remove it. One can toggle dev tools with hotkeys, as opposed to programmatically. Check it out, I instigated a change in the electron repo: https://github.com/atom/electron/issues/2205
:D

For reference, devtools/inspector can be toggled with ctrl+shift+i

app/js/UI.js
Mingling94, line 49: ^yup, I'm going to make a settings page to change these things in session after a few more important features are developed

app/js/daemonManager.js
Mingling94, line 69: Oh true, this JS file is already making a log file!

Mingling94:
Putting this here for reference: I'm moving to this standard of commenting. I've been floundering about to have my own standard, but this one comes with an auto-documentation tool similar to godoc
http://usejsdoc.org/about-getting-started.html

app/css/general.css
```diff
@@ -27,16 +40,29 @@ html,body {
...
-}
-.logo {
 	opacity: 0.7;
 
 	pointer-events: none;
 }
 
+/* Header Buttons */
+.header .button-container {
+}
```
DavidVorick: this css field is empty

app/js/UI.js
DavidVorick, line 49: settings page should wait until the next PR

DavidVorick:
Okay, looks pretty good to me. Is there anything else you want to add before it gets merged?

Mingling94:
Some small corrections, but now it's good! I'll do small features tomorrow like the settings page and the update button

Issue #18, opened 2015-07-19
Overall UI thoughts


DavidVorick:
.gitignore:

`*.swp` instead of `**/*.swp`. Also, it's a bit disorganized, there's probably a better way to sort things.

---

DEVELOPERS.md:

It's probably good enough to call it Developers.md

"To get started, run 'make dependencies && make"

Ming's brief guide to javascript is unnecessary. It would instead be sufficient to have a section titled "Technologies" or something similar to that, and then talk about using node.js and electron (perhaps with a brief explanation for why they were chosen), along with links to learn more about each. Links to javascript style guides and best practices are also okay, but you don't need much more than links unless there's something the project enforces that isn't really idiomatic.

I think for the most part developers.md just needs to be rewritten.

Developers.md should probably explain something of the general project layout, including stuff like plugin information, sidebar information, dependency handling, and topbar information. Basically, enough information that someone can dive right into what they are most interested in working on without needing to look at a huge portion of the project.

---

README.md:

readme looks fine, just remember that it's targeted at users who want to compile things themselves, maybe monkey around with the CSS a bit, but it's not targeted at developers.

I think it's good though right now.

---

package.json:

the "scripts" seem a bit cluttered/disorganized. Otherwise it all looks pretty good. As the project matures further, the scripts should probably be cleaned up, but for now it's fine.

---

app/index.js:

line 24, `appIcon.setToolTip`: can you explain that further? What does it do, and should be change the words?

line 41: `mainWindow.focus()`: is there a specific reason that the UI needs to grab focus? I generally find it annoying when my apps try to grab focus after I open them. I'm typically multitasking.

---

app/index.html:

Is there any reason that the javascript is included at the bottom of the file?

The comment `save the config as is` is a bit misleading. The function is `UI.close`, and while it does save the config, index.html is really the wrong place to say that. The two comments about the UI startup and closing are probably not needed at all, it's pretty clear what's happening in the code.

---

app/assets/logo:

I find it interesting that we have the logo in both a png and an svg. Is there any reason that we need both?

---

app/css/general.css:

don't import fonts via a url, have them downloaded and in the assets folder. We can't assume that the user will always have an internet connection when working in Sia, plus it causes huge latencies that aren't necessary. (this is specifically talking about line 12)

Assuming that there will be at most 5 sidebar buttons is bad. I think 8 is a safer starting point, but as people start making their own plugins that number might go up substantially.

---

app/plugins/Overview:

[...]

Issue #14, opened 2015-07-07
Heavily Documented Example Plugin


DavidVorick:
The overview plugin should be our example plugin. It should be heavily documented, with every bit of code being explained in the context of how you make a plugin. We should expect that people will regularly be using the overview plugin as a reference, and so the code should be fully explained and of the highest quality.

Once the overview plugin is ready, we'll open source the test-ui.

I am worried because the API isn't stable yet, but every day we get closer to refactoring the api.

Mingling94:
API's stable!

I was going to write test cases for all the API calls, though I'll hold off on that because you mentioned overhauling the API which could nullify a lot of call-specific javascripting

DavidVorick:
That's what I mean by a stable API. We're overhauling it, therefore it's not stable.

PR #20, opened 2015-07-20
Documentation

Mingling94:
Moved the plugins back into a plugins folder, amended documentation to address #18
#14 isn't completely solved because I want to approach this using JSDoc 3 standards.

https://github.com/Mingling94/test-ui/blob/982d8c2c18bc1ed7d6f8e9e69a703a179055cdc8/doc/plugin-tutorial.md

Checkout the tutorial :D


DavidVorick:
I looked at the tutorial first, looks pretty good. I think at the end you should display the whole overview.js file so people can see it's completed form. I believe you show the entirety of the other files.

DavidVorick:
``` js
// Variables to store API call values
var calls = ['/wallet/status', '/gateway/status', '/consensus/status'];

function callAPI() {
    calls.forEach(function(call) {
        IPC.sendToHost('api-call', call);
    });
}
```

This portion is a bit sloppy. Namely, the word 'call' appears wayyy too many times. Also, I would think that you would want the `calls` variable to be inside of the function, is there a reason it isn't?

app/js/daemonManager.js
```diff
@@ -45,17 +22,13 @@ var Daemon = (function() {
...
+		if (!isRunning) {
+			isRunning = function() {};
 		}
-		if (typeof(isNotRunning) !== 'function') {
-			isNotRunning = function() {
-				console.log('siad is not running');
-			};
+		if (!isNotRunning) {
+			isNotRunning = function() {};
 		}
```
DavidVorick, line 39: @lukechampine might have some advice

are these checks necessary? I think it's safe to assume that the caller didn't use null.

And, since you DO use null, it's probably better to just send in an empty function.

DavidVorick, line 39: I'm also inclined to say that this abstraction is unnecessary, how often do you call `ifSiad`?

app/js/daemonManager.js
```diff
@@ -84,45 +57,30 @@ var Daemon = (function() {
...
 		console.log('stopping siad');
-		ifSiad(function() {}, function() {
+		ifSiad(null, function() {
 			console.err('attempted to stop siad when it was not running');
 			return;
 		});
-		APIJS.getCall(address + '/daemon/stop', function(err, data) {
+		apiCall('/daemon/stop', function(err, data) {
 			console.assert(!err && data);
 			console.log(data);
```
DavidVorick, line 80: what happens if the daemon is already not running?

DavidVorick, line 80: the user could kill it from the task manager, for example.

app/js/plugin.js
```diff
@@ -10,34 +7,20 @@ var Factory = require('./pluginFactory');
...
 
 	// show() shows the plugin's view
 	function show() {
 		button.classList.add('current');
-		view.classList.remove('hidden');
-		view.classList.add('current');
-		//view.style.display = '';
-		view.send('show');
+		view.style.display = '';
+		view.executeJavaScript('if (typeof init === "function") init();');
```
DavidVorick, line 16: the `if` statement seems weird.

I guess it makes sense if you're trying to be paranoid about people writing plugins incorrectly.

app/js/plugin.js
```diff
@@ -56,7 +39,8 @@ module.exports = function plugin(plugPath, name) {
 	// sendIPC is for communicating with the plugin's webview, while still
 	// keeping it private to plugin
 	function sendIPC(channel, args) {
-		view.send(channel, args);
+		view.send.apply(view, [].slice.call(arguments));
+		//view.send(channel, args);
```
DavidVorick: what was the reason for changing this?

app/js/pluginFactory.js
```diff
@@ -24,16 +25,28 @@ module.exports = {
...
+		v.autosize = 'on';
 		
 		// Turn nodeintegration on so plugins can use electron & node libraries
+		// TODO: nodeintegration breaks some global variables on sites, like
+		// jquery. We should turn nodeintegration on only for the plugins that
+		// need it.
 		v.nodeintegration = 'on';
 		// TODO: If security becomes a concern, specify preload script to give
 		// plugins their dependencies and access to libraries, but turn off
 		// nodeintegration superpowers
```
DavidVorick: neither of these TODOs is really a concern.

app/js/daemonManager.js
Mingling94, line 39: A few times, starting and stopping uses it for safety. Otherwise, other modules might be introduced which could use this. 

Honestly it's just to save me from writing code to send a test call 3x over.

As for the null --> empty function checks, why not?

app/js/daemonManager.js
Mingling94, line 80: The check a few lines up using ifSiad should already prevent stopping a daemon that isn't running.

app/plugins/Overview/index.html
```diff
@@ -2,23 +2,31 @@
 <html>
 	<head>
 		<title>Overview</title>
```
DavidVorick: does having a title make sense?

app/js/plugin.js
Mingling94, line 16: Paranoid, maybe. This really just prevents error output from each plugin that lacks an init/kill function. No need to force these on the plugins, right?

app/plugins/Overview/index.html
DavidVorick: ditto for the tutorial

app/plugins/Overview/js/overview.js
```diff
@@ -0,0 +1,72 @@
...
+// Keeps track of if the view is shown
+var updating;
+// Variables to store API call values
+var calls = ['/wallet/status', '/gateway/status', '/consensus/status'];
+
+function callAPI() {
+	calls.forEach(function(call) {
+		IPC.sendToHost('api-call', call);
+	});
+}
```
DavidVorick: this function just seems like it could be better.

app/plugins/Overview/js/overview.js
```diff
@@ -0,0 +1,72 @@
...
+			blockHeight = result.Height || blockHeight;
+			document.getElementById('block-height').innerHTML = 'Block Height: ' + blockHeight;
+			break;
+		default:
+			console.error('Unexpected call: ' + call + ' and result: ' + result);
+	}
+}
+
+// Update values per call
+IPC.on('api-result', update);
```
DavidVorick: This file has no docstrings

app/js/plugin.js
Mingling94: This one's an interesting JS thing. Calling `anyFunction.apply(this_param, array)` calls the function, giving this_param as the `this` value and an array to allow a variable amount of arguments. I encountered problems sending, not just the callresult from an API call, but the original call string/object as well as a possible error. I would have to send a specific number of arguments otherwise.

In fact sendIPC shouldn't even have parameters anymore.

app/js/daemonManager.js
DavidVorick, line 39: it has opaque usage. I'm wondering what Luke's opinion is on this.

3x is a good reason to build an extra function.

app/js/pluginFactory.js
Mingling94: Word, I'll take them out tomorrow.

app/js/plugin.js
DavidVorick, line 16: I guess not.

DavidVorick:
Ok, that sums up my review of this PR. Not too many things.

app/plugins/Overview/js/overview.js
Mingling94: Alternatively, it could be individually called:

``` js
IPC.sendToHost('api-call', '/wallet/status');
IPC.sendToHost('api-call', '/gateway/status');
IPC.sendToHost('api-call', '/consensus/status');
```

Or use the standard for loop.

I could just rename the repetition of the word 'call' out of this. I'm of the opinion `array.forEach(function(element, index, array) { // Executed per array element });` is preferable to for loops in JS

app/plugins/Overview/js/overview.js
Mingling94: Gonna do some JSDoc flexing tomorrow :tongue:

app/plugins/Overview/js/overview.js
DavidVorick: array.forEach is an odd construction to me, I think a for loop would be more familiar to most people.

I'm not too worried about it though.

Issue #15, opened 2015-07-08
Infinite Precision Numbers


DavidVorick:
We need all of the numbers in the UI to have infinite precision. I believe having integers that go up to 128bits is sufficient, but it's probably going to be easier to find an infinite precision library.

There were various problems on the old UI because 1 siacoin is actually 10e24 base units, and because 1 MS is actually 10e30 base units. So if you try sending 1.031 MS, you often end up actually sending 1.031000000000000000001 base units, and this confuses people and shouldn't happen.

PR #21, opened 2015-07-23
Adhered Comments To JSDoc standards

Mingling94:
I should figure how to do this, but putting these changes in one commit is
easier for now

Check out New-Sia-UI/jsdoc/Sia-UI/0.4.0/index.html! The JSDoc folder is an auto-generated documentation site. It probably shouldn't be incorporated in the repo in hindsight, but I'm not sure where to put it

PR #22, opened 2015-07-26
Precision

Mingling94:
Have yet to make unit tests, but I'd want to crank out some plugins first in case I discover the need to redesign something from the process.

Addresses #15


DavidVorick:
this guy needs to be rebased on top of #21

Mingling94:
@DavidVorick Yoyo rebased

PR #24, opened 2015-07-27
Alternate way of calling api from overview plugin

Mingling94:
More abstraction and less use of IPC throughout. :+1: or :-1:?


Mingling94:
#25 more favorable

PR #28, opened 2015-08-03
Wallet (various fixes)

Mingling94:
CSS for wallet was off, didn't notice before last PR.

Boxes are now half the range of the height now and colored differently.

### DIFF
--- a/app/plugins/Wallet/js/wallet.js
+++ b/app/plugins/Wallet/js/wallet.js
@@ -37,6 +37,18 @@ function formatSiacoin(hastings) {
 	return display + " SC";
 }
 
+// Adds a given address to the list
+function appendAddress(address) {
+	if (eID(address)) {
+		return;
+	}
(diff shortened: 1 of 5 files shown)


Mingling94:
@DavidVorick boop, this one's probably more important than the other PR

app/plugins/Wallet/js/wallet.js
```diff
@@ -115,17 +127,10 @@ function update() {
 			return;
 		}
 		
+		console.log('updating...');
```
lukechampine: debugging output?

lukechampine:
a lot of the CSS indentation seems off

Mingling94:
Blast! I should just set tabbing to be something weird like 6 characters so its obvious when I'm actually looking at 4 spaces

Mingling94:
@lukechampine Applied changes to your comments ^_^

PR #39, opened 2015-08-11
Reorg

Mingling94:
Due to the proposed changes to wallet, plugin IPC has to be done a bit differently.

Plugins now determine a channel for IPC to listen on and respond with when making a call. This is because the call url no longer can serve appropriately as a response IPC channel when multiple API functions are served on the same endpoint.

Also moved index.js to root level and linted.

@lukechampine @DavidVorick This PR is important for how I proceed with the hosting plugin

TODO: after this PR is accepted, reflect it in the plugin tutorial.

Leftover TODO: from a previous PR, reflect notification and tooltip system in plugin tutorial.

### DIFF
--- a/app/plugins/Overview/js/overview.js
+++ b/app/plugins/Overview/js/overview.js
@@ -21,7 +21,7 @@ function update() {
 function updateField(err, caption, newValue, elementID) {
 	if (err) {
 		console.error(err);
-	} else if (!newValue) {
+	} else if (newValue === null) {
 		console.error('Unknown occurence: no error and no result from API call!');
 	} else {
 		document.getElementById(elementID).innerHTML = caption + newValue;
--- a/app/plugins/Overview/js/overview.js
+++ b/app/plugins/Overview/js/overview.js
@@ -3,25 +3,27 @@
 const IPC = require('ipc');
 // Library for arbitrary precision in numbers
 const BigNumber = require('bignumber.js');
+// Ensure precision
+BigNumber.config({ DECIMAL_PLACES: 24 });
+BigNumber.config({ EXPONENTIAL_AT: 1e+9 });
 // Keeps track of if the view is shown
 var updating;
-// Keeps track of if listeners were already instantiated
-var listening = false;
 
-// Call API and listen for response to call
-function callAPI(call, callback) {
-	IPC.sendToHost('api-call', call);
-	// prevents adding duplicate listeners
-	if (!listening) {
-		IPC.on(call, function(err, result) {
-			if (err) {
-				console.error(err);
-			} else if (result) {
-				callback(result);
-			} else {
-				console.error('Unknown occurence: no error and no result from callAPI!');
-			}
-		});
+// Make API calls, sending a channel name to listen for responses
+function callAPI() {
+	IPC.sendToHost('api-call', '/wallet/status', 'balance');
+	IPC.sendToHost('api-call', '/gateway/status', 'peers');
+	IPC.sendToHost('api-call', '/consensus/status', 'height');
+}
+
+// Updates element text
+function updateField(err, caption, newValue, elementID) {
+	if (err) {
+		console.error(err);
+	} else if (!newValue) {
+		console.error('Unknown occurence: no error and no result from callAPI!');
+	} else {
+		document.getElementById(elementID).innerHTML = caption + newValue;
 	}
 }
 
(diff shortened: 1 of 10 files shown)


app/plugins/Overview/js/overview.js
```diff
@@ -3,25 +3,28 @@
 const IPC = require('ipc');
 // Library for arbitrary precision in numbers
 const BigNumber = require('bignumber.js');
+// Ensure precision
+BigNumber.config({ DECIMAL_PLACES: 24 });
+BigNumber.config({ EXPONENTIAL_AT: 1e+9 });
```
DavidVorick, line 8: Can you explain the reasoning for the exponential at 1e9? Will it ever even be displayed this way?

I guess it'll make more sense once I see it in action

app/plugins/Overview/js/overview.js
```diff
@@ -3,25 +3,28 @@
...
-				callback(result);
-			} else {
-				console.error('Unknown occurence: no error and no result from callAPI!');
-			}
-		});
+// Make API calls, sending a channel name to listen for responses
+function update() {
+	IPC.sendToHost('api-call', '/wallet/status', 'balance');
+	IPC.sendToHost('api-call', '/gateway/status', 'peers');
+	IPC.sendToHost('api-call', '/consensus/status', 'height');
```
DavidVorick: little bit weird to name them after the fields you are pulling from them, instead of naming them after the logic.

I'd probably pick names like 'walletStatus' or something.

app/plugins/Wallet/js/wallet.js
```diff
@@ -3,34 +3,17 @@
 const IPC = require('ipc');
 // Library for arbitrary precision in numbers
 const BigNumber = require('bignumber.js');
-// Variables to store api result values
+// Ensure precision
+BigNumber.config({ DECIMAL_PLACES: 24 });
+BigNumber.config({ EXPONENTIAL_AT: 1e+9 });
```
DavidVorick, line 8: maybe these should be a globally accessible variable.

maybe not

app/plugins/Wallet/js/wallet.js
```diff
@@ -109,61 +81,83 @@ function verifyTransaction(callback) {
...
-// Called upon showing
-function init() {
-	// DEVTOOL: uncomment to bring up devtools on plugin view
-	// IPC.sendToHost('devtools');
-	
-	// Ensure precision
-	BigNumber.config({ DECIMAL_PLACES: 24 });
-	BigNumber.config({ EXPONENTIAL_AT: 1e+9 });
+};
+eID('confirm').onclick = function() {
```
DavidVorick, line 108: what is `eID`?

app/plugins/Overview/js/overview.js
Mingling94, line 8: From the BigNumber documentation:

> EXPONENTIAL_AT
> number: integer, magnitude 0 to 1e+9 inclusive, or 
> number[]: [ integer -1e+9 to 0 inclusive, integer 0 to 1e+9 inclusive ]
> Default value: [-7, 20]
> The exponent value(s) at which toString returns exponential notation.
> 
> ``` js
> // Almost never return exponential notation:
> BigNumber.config({ EXPONENTIAL_AT: 1e+9 })
> ```

app/plugins/Wallet/js/wallet.js
Mingling94, line 8: There's no way to make the UI's global variables available to the plugins. Since it's across a different index.html and functionally a different webpage being displayed, it has a different global scope than the UI's global scope.

Unless you mean it should be in the plugin's index.html

app/plugins/Wallet/js/wallet.js
Mingling94, line 108: It's at the top and is just a shorthand for `document.getElementById()`

``` js
function eID() {
    return document.getElementById.apply(document, [].slice.call(arguments));
};
```

app/plugins/Overview/js/overview.js
Mingling94: I'm going to stick to hyphen separated terms like 'new-address' or 'balance-update' instead of camelCase because that's generally a variable & function naming convention.

Mingling94:
boop @DavidVorick

app/plugins/Wallet/js/wallet.js
DavidVorick, line 108: hmm. Not quite clear, but I guess it's okay.

PR #41, opened 2015-08-12
Siadhandling

Mingling94:
`plugin.show()` and `plugin.hide()` now calls, via IPC, `view.start()` and `view.stop()` for semantics and to differentiate between the two. `plugin.show()` is called upon the button being pressed and incorporates transition logic, `view.start()` is called to start updating and api calling.

This PR makes siad startup handling a bunch more explicit, sending notifications to the user upon the daemon starting, stopping, and exiting. A new public boolean, named, "Running" in the DaemonManager keeps track of if the DaemonManager is aware that siad has started. Prevents an infinite flow of failed API calls when Running is set to false.

TODO after PR accepted: prettify notifications and allow user to turn siad on and off

### DIFF
--- a/app/plugins/Overview/index.html
+++ b/app/plugins/Overview/index.html
@@ -3,6 +3,7 @@
 	<head>
 		<!-- CSS -->
 		<link rel="stylesheet" href="../../css/roboto-condensed-min.css">
+		<link rel="stylesheet" href="../../css/font-awesome-4.3.0/css/font-awesome.min.css">
 		<link rel="stylesheet" href="../../css/plugin-standard.css">
 		<link rel="stylesheet" href="css/overview.css">
 	</head>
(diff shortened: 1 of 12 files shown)


Mingling94:
boop @lukechampine @DavidVorick

lukechampine:
why would the user want to turn siad on and off?

app/plugins/Overview/index.html
```diff
@@ -22,6 +23,10 @@
 			<div class="welcome">
 				<div class="large">Welcome to Sia</div>
 				<div class="small">A highly efficient decentralized storage network.</div>
+				<div class="button" id="stop-exit">
+					<i class="fa fa-sign-out"></i>
+					Stop siad & exit
```
lukechampine: needs proper escaping (`&amp;`)

Mingling94:
@lukechampine This only lets the user turn siad off. The UI turns siad on startup. @DavidVorick suggested that there should be a siad on/off button but it is an odd use case I suppose.

Mingling94:
@lukechampine I took out the stop siad thing in overview as well as in daemonManager until these use cases become a bit more clear.

lukechampine:
needs rebase

Mingling94:
@lukechampine rebased

PR #42, opened 2015-08-12
Doc

Mingling94:
Documents the changes made in #39 in plugin-tutorial.md. Made on top of #41


app/js/daemonManager.js
```diff
@@ -20,6 +20,18 @@ function DaemonManager() {
...
+	/**
+	 * Empty function to use with ifSiad()
+	 * @member {function} DaemonManager~EMPTYFUNC
+	 */
+	var EMPTYFUNC = function() {};
+	/**
+	 * Boolean to track if siad is running
+	 * @member {function} DaemonManager.Running
+	 */
+	this.Running = false;
```
lukechampine, line 27: that doesn't look like a function...wrong docstring?

doc/plugin-tutorial.md
```diff
@@ -341,70 +341,101 @@ the UI reacts accordingly. For example, turn chromium devtools on using:
...
+```
+
+* Note, a call object is required instead of a string url for any calls other
+than 'GET'. They are formatted like the following:
+
+```json
+{ // A call object
+	url: '/consensus/status',  // String
+	type: 'GET',               // String
+	args: null,                // null or an object containing arguments to send
```
lukechampine, line 381: comments aren't allowed in json. Try changing the language to js.

lukechampine:
needs rebase

Mingling94:
@lukechampine rebased

doc/plugin-tutorial.md
```diff
@@ -675,35 +655,80 @@ function formatSiacoin(hastings) {
...
+// Give the buttons interactivity
+document.getElementById('create-address').onclick = function() {
+	tooltip('Creating...', this);
+	IPC.sendToHost('api-call', '/wallet/address', 'new-address');
+};
 ```
+
+Which will produce this tooltip on click appear as such:
+
+![It's the details that make the best UI] (/doc/assets/wallet-tooltip.png)
```
lukechampine: this image (and the one above) are showing up as links in the preview. There isn't usually a space between the `]` and `(`, maybe that's the problem. Otherwise it could just be a diff rendering quirk.

Mingling94: It definitely is the space

PR #44, opened 2015-08-15
add information about wallet locking to the overview

DavidVorick:
pretty simple


Mingling94:
Checked it, looks good!

PR #57, opened 2015-08-18
Walletfixes

Mingling94:
On top of #55 

Transaction history needed a touchup to display its transactions well

### DIFF
--- a/app/plugins/Wallet/js/helpers.js
+++ b/app/plugins/Wallet/js/helpers.js
@@ -34,6 +34,7 @@ function hide(el) {
 }
 
 // Convert to Siacoin
+// TODO: Enable commas for large numbers
 function convertSiacoin(hastings) {
 	var ConversionFactor = new BigNumber(10).pow(24);
 	var display = new BigNumber(hastings).dividedBy(ConversionFactor);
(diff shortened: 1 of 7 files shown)


app/plugins/Wallet/js/helpers.js
```diff
@@ -5,7 +5,7 @@ const IPC = require('ipc');
 // Library for arbitrary precision in numbers
 const BigNumber = require('bignumber.js');
 // Ensure precision
-BigNumber.config({ DECIMAL_PLACES: 24 });
+BigNumber.config({ DECIMAL_PLACES: 30 });
```
DavidVorick, line 8: This change keeps full precision out to the millions?

Seems odd that it's a parameter at all? Shouldn't you just keep an infinite precision int in the background and display a truncated one to the users?

DavidVorick:
despite having open comments, there's nothing here that's worth holding back merging, as far as I can tell.

app/plugins/Wallet/js/helpers.js
Mingling94, line 8: My thought was the largest unit is a MegaSiacoin via the send transaction so someone would be able to send 1.000000000000000000000000000001 MS to send 1 MS + 1 hasting without rounding

DavidVorick, line 8: I don't think that's a realistic use case, but I guess it's good to support it, because sometimes people do weird things.

PR #61, opened 2015-08-22
Misc. Fixes

Mingling94:
OSX usability without colliding with hotkeys of other OSes

Close the app on all-windows-closed on OSX as suggested by chanc3r on slack

Added unlocking spinner

Fixed plugin-standard css a bit


Mingling94:
@DavidVorick This should be reviewed.

I responded to a few usability concerns on slack with this PR

DavidVorick:
Great, thanks. Looking at both the renter PR and this PR as soon as possible.

Glad to have this finally coming together. We are in the final stages of testing `siad` and `siac`, as soon as that's complete I'll turn my full attention to the UI until it's ready to launch.

DavidVorick:
Tried it out, get a constant stream of 'siad seems to have stopped working' messages after killing siad

edit: seems like they eventually stop, but only after close to 100 of them appear

Mingling94:
Hmm, haven't encountered that. Do you start the UI with siad already running?

Also this is on Overview plugin, yes?

Mingling94:
Also I'm not sure how this could be occurring. pluginManager only sends calls `if (Daemon.Running)` and `if (err)`, it checks `ifSiad()` which, `if (err)` again, is setting `self.Running = false;`

Edit, prematurely closing siad while the UI is running on Overview gives me 3 of such errors because I guess all 3 occur simultaneously and thus none of them have daemonManager's `self.Running = false;` by the time they error. However for something like a 100 to happen at once...

Mingling94:
Pretty sure I've now dealt with most all of [gabes's report](https://siatalk.slack.com/files/gabes/F09JKFSDP/Sia-UI-linux-x64_0_4_0_Bugreport)

Two things that are above me:
- [Electron's accelerator errors](http://pastebin.com/shTFgjNY)
- > Wallet is locked, popup message: error 404 Not found 404 - Refer to API.md ( delete wallet json and restart app doesn't help)

PR #64, opened 2015-08-27
Adjusted api calls to use string numbers

Mingling94:
@DavidVorick You said Sia master had merged something that changed number returns from API to strings? I don't see that on the Sia repo


DavidVorick:
Is this PR working with the recent 'siad'? I'm pretty sure the code changed at some point. Perhaps in Luke's 85 commit PR?

Mingling94:
Luke's PR seemed to not change anything about the host api that would change the UI, correct me if I'm wrong.

The renter no longer has a 'Repairing' field to its file list, that's alright I responded to that in my renting branch

PR #67, opened 2015-09-03
Wallet address list and v0.3.3.x loading

lukechampine:
This would have been really easy if only loading old wallets didn't require the encryption password. As a result my implementation is slightly hacky. That's sort of the name of the JavaScript game though.


DavidVorick:
Tried it out, looks good, seems to work smoothly.

Mingling94:
So why not have a separate popup instead of reusing the password one?

DavidVorick:
I think it was just easier to program that way.

PR #68, opened 2015-09-05
Build

DavidVorick:
Sia building now works.

I couldn't get windows branding working from my computer, wine was having troubles. So I left it out. But the mac rebranding and linux rebranding should be complete.

I pulled a few objects (such as the logo, and the Sia folder) up into a higher-level directory. I also removed the absolute paths from the config file.


DavidVorick:
this never got merged?

app/js/daemonManager.js
```diff
@@ -73,7 +73,7 @@ function DaemonManager() {
 			if (err) {
 				self.Running = false;
 				// Check again later
-				setTimeout(updatePrompt, 1000);
+				setTimeout(updatePrompt, 100000); // Not great - we'll have to fix that.
```
lukechampine: why extend this?

DavidVorick: goal was to wait until the wallet had finished unlocking, according to a comment. 1 second is definitely not enough even on my desktop.

but maybe 100 seconds is too many. I'm not even sure it does anything.

DavidVorick: ohh I think I completely misunderstood what was happening here. 1 second is probably fine.

app/plugins/Files/js/helpers.js
```diff
@@ -3,7 +3,7 @@
 // Library for communicating with Sia-UI
 const IPC = require('ipc');
 // Library for arbitrary precision in numbers
-const BigNumber = require('bignumber.js');
+const BigNumber = require('../../js/bignumber.min.js');
```
Mingling94, line 6: Err, why?

package.json
```diff
@@ -1,12 +1,11 @@
...
-  "version": "0.4.1",
-  "main": "index.js",
+  "name": "Sia-UI-beta",
+  "version": "0.4.2",
+  "main": "app/index.js",
   "author": "NebulousLabs",
   "description": "A UI application for interfacing with Sia",
   "license": "MIT",
   "devDependencies": {
-    "electron-packager": "latest",
```
Mingling94, line 9: I'm a little against this but it's out of my hands now. Does the app icon and tray icon still work?

package.json
```diff
@@ -15,33 +14,22 @@
...
-    "build-win32": "npm run build-cmd -- --platform=win32 --arch=ia32",
-    "build-mac": "npm run build-cmd -- --platform=darwin --arch=x64",
-    "build-cmd": "electron-packager . Sia-UI --version=0.30.0 --prune --ignore='doc|build' --out=build --icon=app/assets/logo/sia.svg"
+    "lint": "jshint . --verbose --exclude-path=.jshintignore",
+    "package": "make package",
+    "package-linux64": "make package-linux64",
+    "package-linux32": "make package-linux32",
+    "package-win64": "make package-win64",
+    "package-win32": "make package-win32",
+    "package-osx": "make package-osx"
```
Mingling94: Oh, well I guess that works. Why not describe these npm scripts in the Developers.md instead so as to avoid instructions about both `npm run scripts` and `make scripts`?

app/index.html
```diff
@@ -45,18 +45,18 @@
...
+			const ElectronScreen = require('screen');
 			const Process = require('child_process').spawn;
-			const Path = require("path");
-			const Fs = require("fs");
-			const Shell = require("shell");
-			const $ = require("jquery");
+			const Path = require('path');
+			const Fs = require('fs');
+			const Shell = require('shell');
+			const $ = require('./js/jquery.min.js');
```
Mingling94, line 55: Why a minified js file instead of the node package? Same principle as BigNumber I guess. You can avoid dependency's updates breaking your usage by specifying a version number in the package.json instead of `latest` ya know.

Makefile
```diff
@@ -0,0 +1,115 @@
+# This Makefile should really be a bash script, but I don't know how to make
+# bash scripts that have multiple options, and since we're under deadline
+# pressure I've elected not to learn.
```
Mingling94: You would just have multiple bash scripts and choose one to run. Otherwise you would capture a cli argument and run segments based on the argument.

app/plugins/Files/js/helpers.js
DavidVorick, line 6: didn't have nodejs on my computer, couldn't run electron without it

package.json
DavidVorick, line 9: electron packager was putting things together wrong, and doing absurd things like resources/app/app/dependencies/Sia

package.json
DavidVorick: I only used a makefile because I don't know to write bash scripts. But more standard practise is to write some bash scripts and run them from node.

app/index.html
DavidVorick, line 55: It was more about running into errors that were preventing me from running it.

Makefile
DavidVorick: I don't know how to write bash lol

app/plugins/Files/js/helpers.js
Mingling94, line 6: Didn- DIDN'T HAVE NODEJS? It's a Nodejs project -____-
Hope it's not too limiting/messy in the future...

package.json
Mingling94: 1. Makefiles and bash are quite similar
2. Having the documentation (Developers.md) reference only npm commands would be more coherent is allsImsayin

app/plugins/Files/js/helpers.js
Mingling94, line 6: You don't find it odd that you're using _node package manager_ without nodejs?

DavidVorick, line 6: I will confess it's not the best plan in the world.  I still haven't fully embraced the node way yet.

Things will continue to evolve, but I needed to be able to build the app in the meantime.

package.json
DavidVorick: yeah we'll get there.

DavidVorick: wait, doesn't Developers.md already only reference npm commands?

doc/Developers.md
```diff
@@ -64,4 +64,10 @@ this UI project (e.g. image logos, third-party css, etc.)
 Lastly, the 'dependencies' folder contains anything that are not a part of the
 project, but that it needs to use. Currently this is only siad.
 
+## Packaging Sia-UI
 
+There is a Makefile that contains some packaging scripts. Run `make package` to
+package Sia. You will need to have the primary Sia project in your GOPATH, and
+you will need to have run `make xc` from that project. Running `make package`
+will grab the packaged binaries from the 'Sia' repo and use them to package the
+Sia-UI.
```
Mingling94: Here is Developers.md referencing the Makefile

DavidVorick: ah ok, I can correct this.

Makefile
lukechampine: probably a good thing to learn if you plan on using Linux for the rest of your life

lukechampine:
I made a bash script.

DavidVorick:
I haven't tested it, but as long as it works you can remove the Makefile

DavidVorick:
I merged it perhaps prematurely, but it works for me and I wanted to tag + release

PR #76, opened 2015-09-22
Add option to upload a directory

pmknutsen:
- Selected directory is uploaded single-file, non-recursively.
- Files can be given a prefix (default is directory name).
- Hidden files are ignored.
- Minor CSS fix to ensure input fields fit inside window.

### DIFF
--- a/app/plugins/Files/index.html
+++ b/app/plugins/Files/index.html
@@ -80,6 +80,19 @@
 					<i class='fa fa-close'>Cancel</i>
 				</span>
 			</div>
+			<!-- Delete Popup -->
+			<div class='popup hidden' id='confirm-delete'>
+				<div class='row'>
+					<div class='title'></div>
+				</div>
+				<p class='nickname'></p>
+				<span class='button' id='delete-file'>
+					<i class='fa fa-check'>Delete</i>
+				</span>
+				<span class='button' id='cancel-delete'>
+					<i class='fa fa-close'>Cancel</i>
+				</span>
+			</div>
 		</div>
 		<!-- Sliding New File Frame -->
 		<div class='frame hidden' id='add-file'>
--- a/app/plugins/Files/index.html
+++ b/app/plugins/Files/index.html
@@ -47,13 +47,13 @@
 					<div class='pure-u-1-2'>
 						<div class='button' id='new-file'>
 							<i class='fa fa-plus'></i>
-							Add file
+							Add File
 						</div>
 					</div>
 					<div class='pure-u-1-2'>
 						<div class='button' id='new-dir'>
-							<i class='fa fa-plus'></i>
-							Add files in directory
+							<i class='fa fa-folder-open'></i>
+							Add Folder
 						</div>
 					</div>
 					<!-- TODO: Search Bar -->
@@ -138,33 +138,33 @@
 				</div>
 			</div>
 		</div>
-		<!-- Sliding New Directory Frame -->
+		<!-- Sliding New Folder Frame -->
 		<div class='frame hidden' id='add-dir'>
 			<div class='header'>
 				<div class='button' id='back-dir'>
 					<i class='fa fa-chevron-left'></i>
 				</div>
 				<div class='title'>
-					Upload Directory
+					Add Folder
 				</div>
 			</div>
 
-			<div class='pure-g step' id='step1'>
-				<div class='pure-u-1-3 button' id='upload-dir-choice'>
+			<div class='pure-g step' id='step11'>
+				<div class='pure-u-1-1 button' id='upload-dir-choice'>
 					<i class='fa fa-upload'></i>
-					<p>Select Directory</p>
+					<p>Select Folder</p>
 				</div>
 			</div>
 
-			<div class='pure-g step' id='step2'>
+			<div class='pure-g step' id='step21'>
 				<div class='pure-u-1-1 hidden' id='nickname-dir'>
-					<p>Set prefix of uploaded files</p>
+					<p>Set optional file prefix</p>
 					<input class='description-field' id='nickname-dir-input'></input>
 					<div class='dir-path hidden'></div>
 				</div>
 			</div>
 
-			<div class='pure-g step' id='step3'>
+			<div class='pure-g step' id='step31'>
 				<div class='pure-u-1-1 hidden button' id='upload-dir'>
 					<i class='fa fa-plus'></i>
 					<p>Upload</p>
--- a/app/plugins/Files/index.html
+++ b/app/plugins/Files/index.html
@@ -83,7 +83,7 @@
 			<!-- Delete Popup -->
 			<div class='popup hidden' id='confirm-delete'>
 				<div class='row'>
-					<div class='title'></div>
+					<div class='title'>Confirm</div>
 				</div>
 				<p class='nickname'></p>
 				<span class='button' id='delete-file'>
--- a/app/plugins/Files/index.html
+++ b/app/plugins/Files/index.html
@@ -149,14 +149,14 @@
 				</div>
 			</div>
 
-			<div class='pure-g step' id='step11'>
+			<div class='pure-g step' id='stepf1'>
 				<div class='pure-u-1-1 button' id='upload-dir-choice'>
 					<i class='fa fa-upload'></i>
 					<p>Select Folder</p>
 				</div>
 			</div>
 
-			<div class='pure-g step' id='step21'>
+			<div class='pure-g step' id='stepf2'>
 				<div class='pure-u-1-1 hidden' id='nickname-dir'>
 					<p>Set optional file prefix</p>
 					<input class='description-field' id='nickname-dir-input'></input>
@@ -164,7 +164,7 @@
 				</div>
 			</div>
 
-			<div class='pure-g step' id='step31'>
+			<div class='pure-g step' id='stepf3'>
 				<div class='pure-u-1-1 hidden button' id='upload-dir'>
 					<i class='fa fa-plus'></i>
 					<p>Upload</p>
--- a/app/plugins/Files/index.html
+++ b/app/plugins/Files/index.html
@@ -47,7 +47,13 @@
 					<div class='pure-u-1-2'>
 						<div class='button' id='new-file'>
 							<i class='fa fa-plus'></i>
-							New File
+							Add file
+						</div>
+					</div>
+					<div class='pure-u-1-2'>
+						<div class='button' id='new-dir'>
+							<i class='fa fa-plus'></i>
+							Add files in directory
 						</div>
 					</div>
 					<!-- TODO: Search Bar -->
@@ -89,7 +95,7 @@
 			<div class='pure-g step' id='step1'>
 				<div class='pure-u-1-3 button' id='upload-choice'>
 					<i class='fa fa-upload'></i>
-					<p>Upload file</p>
+					<p>Select File</p>
 				</div>
 				<div class='pure-u-1-3 button' id='sia-choice'>
 					<i class='fa fa-file-o'></i>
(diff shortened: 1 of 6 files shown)


DavidVorick:
I haven't read the code yet but..... wow, thanks! Will look at it more over the next few days.

app/plugins/Files/index.html
```diff
@@ -47,7 +47,13 @@
...
 						<div class='button' id='new-file'>
 							<i class='fa fa-plus'></i>
-							New File
+							Add file
+						</div>
+					</div>
+					<div class='pure-u-1-2'>
+						<div class='button' id='new-dir'>
+							<i class='fa fa-plus'></i>
+							Add files in directory
```
Mingling94: Wouldn't this more conventionally be "Add Folder"?

app/plugins/Files/index.html
```diff
@@ -47,7 +47,13 @@
 					<div class='pure-u-1-2'>
 						<div class='button' id='new-file'>
 							<i class='fa fa-plus'></i>
-							New File
+							Add file
+						</div>
+					</div>
+					<div class='pure-u-1-2'>
+						<div class='button' id='new-dir'>
+							<i class='fa fa-plus'></i>
```
Mingling94: Perhaps use another icon, just for less redundancy (two pluses would be odd)

app/plugins/Files/index.html
```diff
@@ -132,6 +138,39 @@
...
+				<div class='button' id='back-dir'>
+					<i class='fa fa-chevron-left'></i>
+				</div>
+				<div class='title'>
+					Upload Directory
+				</div>
+			</div>
+
+			<div class='pure-g step' id='step1'>
+				<div class='pure-u-1-3 button' id='upload-dir-choice'>
```
Mingling94: `pure-u-1-3` was just chosen for the "Add File" frame because there were three options. This looks like it would just be 1 option with 2/3 of the space unoccupied. If you want to keep it as a separate menu, this should be `pure-u-1-1` right?

Otherwise You might also just incorporate it into the other frame and make all the step1 options `pure-u-1-4`

app/plugins/Files/index.html
```diff
@@ -132,6 +138,39 @@
...
+			<div class='header'>
+				<div class='button' id='back-dir'>
+					<i class='fa fa-chevron-left'></i>
+				</div>
+				<div class='title'>
+					Upload Directory
+				</div>
+			</div>
+
+			<div class='pure-g step' id='step1'>
```
Mingling94: HTML element id's should be unique. Now there are two elements with `id='step1'` which would make `document.getElementById('step1')` work improperly. This is especially important with the function called `hideSteps(number)` in buttons.js

app/plugins/Files/js/buttons.js
```diff
@@ -95,6 +97,8 @@ eID('add-sia-file').onclick = function() {
 	loadDotSia(loadPath);
 };
 
+
```
Mingling94: Whattup with the extra spacing?

app/plugins/Files/js/loaders.js
```diff
@@ -59,6 +59,37 @@ addResultListener('uploaded', function(result) {
...
+
+				// Check that file exists
+				var stats = fs.statSync(filePath);
+
+				// Skip hidden files and directories
+				if (~isUnixHiddenPath(filePath) & stats.isFile()) {
+					upload(filePath, nickname + file);
+				}
+			}
+			catch (e) {
```
Mingling94: What error are you catching with this try/catch? Mixing asynchronous error handling with synchronous (try/catch) error handling is a node.js no-no.

Mingling94:
Yo awesome work with this! You understood the code pretty well to do this.

Though I'm curious if and how it works because I'm sketched out by the duplicate html elements with the same id's.

app/plugins/Files/js/loaders.js
pmknutsen: statSync throws an error when file does not exist. Go figure... I did not see how to get around a try/catch here.

pmknutsen:
Thanks, I made changes based on your feedback. HTML id's are fixed. Importantly, hideStep() now takes an array as input (saves a few lines overall). In my tests, the results are variable... Usually, I end up getting many 500 errors thrown by the backend: "Upload failed: not enough hosts to support upload". This happens also when uploading single files. It does not make for a happy user experience... Any ideas?

app/plugins/Files/js/loaders.js
lukechampine: I think you can use `fs.existsSync` to check for existence before calling `fs.statSync`.

If the uploads happened in parallel, you could probably omit the existence check. It looks like they're happening serially here, which is fine for now.

actually, if the file doesn't exist at upload time, `upload` will return an error anyway, so you may not need to do any client-side existence checking whatsoever.

Mingling94: You should use [`fs.stat`](https://nodejs.org/api/fs.html#fs_fs_stat_path_callback) which holds two arguments, a file path and a callback which in turn holds an error and your stats variable, e.g.

```
fs.stat(filePath, function(err, stats) {
    if (err) {
        // handle it
        return;
    }
    // do stuff with stats
}
```

app/plugins/Files/js/buttons.js
```diff
@@ -3,30 +3,32 @@
...
 
-// Used to hide subsequent steps when selecting and earlier one
-function hideSteps(number) {
-	var c = eID('step' + number).children;
-	for (var i = 0; i < c.length; i++) {
-		if(!hidden(c[i])) {
-			hide(c[i]);
+// Used to hide subsequent steps when selecting an earlier one
+function hideSteps(steps) {
+	steps.forEach( function(step) {
```
Mingling94: generally the code base would format this as `steps.forEach(function(step) {`

app/plugins/Files/index.html
```diff
@@ -132,6 +138,39 @@
...
+			<div class='header'>
+				<div class='button' id='back-dir'>
+					<i class='fa fa-chevron-left'></i>
+				</div>
+				<div class='title'>
+					Add Folder
+				</div>
+			</div>
+
+			<div class='pure-g step' id='step11'>
```
Mingling94: This seems messy. How about the ids of the singular add-file steps be `filestep1` and the folder steps be `folderstep1`?

Edit: nevermind, with the `hideSteps()` down below this wouldn't work. If you can't figure a way to work these steps to show up _over_ the original step elements, then this will have to do :D

app/plugins/Files/js/buttons.js
```diff
@@ -127,6 +129,43 @@ eID('cancel-ascii').onclick = function() {
...
+// Select directory sliding frame
+eID('new-dir').onclick = function() {
+	show('add-dir');
+	hide('file-library');
+};
+eID('back-dir').onclick = exitFileAdder;
+
+// Upload directory option chosen
+eID('upload-dir-choice').onclick = function() {
+	var path = require("path");
```
Mingling94: Do not require libraries within commonly occurring events. I forget exactly how, but this causes a mem leak. I took to the convention of requiring libraries at the top of `helpers.js`

app/plugins/Files/js/buttons.js
```diff
@@ -127,6 +129,43 @@ eID('cancel-ascii').onclick = function() {
...
+	});
+	if (loadPath) {
+		eID('nickname-dir').querySelector('.dir-path').innerHTML = loadPath;
+		loadPath = loadPath[0].split(path.sep);
+		eID('nickname-dir-input').value = loadPath[loadPath.length - 1] + '_';
+		show('nickname-dir');
+		show('upload-dir');
+		eID('nickname-dir-input').focus();
+	}
+};
```
Mingling94, line 167: This should exhibit some feedback when the loadPath is wrong, right? That can definitely happen with the Windows file browser at least.

app/plugins/Files/js/buttons.js
```diff
@@ -127,6 +129,43 @@ eID('cancel-ascii').onclick = function() {
...
+
+eID('upload-dir').onclick = function() {
+	var loadPath = eID('nickname-dir').querySelector('.dir-path').innerHTML;
+	var nickname = eID('nickname-dir-input').value;
+	// Illegal filename characters in nickname seems to throw errors
+	// So, substitute \ and / with underscore (_)
+	nickname.replace(/[/\\\\]/g, "_");
```
Mingling94, line 174: It's too early in the morning for me to think of this in-depth, but keep the different OS filename standards in mind.

Mingling94:
Still looks good! As you explore this new option of uploading a directory, feel free to change design and html in large strokes if needed for elegance. Thanks a lot!

app/plugins/Files/js/loaders.js
```diff
@@ -59,6 +59,37 @@ addResultListener('uploaded', function(result) {
...
+		files.forEach(function(file) {
+			try {
+				var filePath = path.join(dirPath, file);
+
+				// Check that file exists
+				var stats = fs.statSync(filePath);
+
+				// Skip hidden files and directories
+				if (~isUnixHiddenPath(filePath) & stats.isFile()) {
+					upload(filePath, nickname + file);
```
Mingling94, line 80: It'd be interesting, yet slightly opaque behavior, to detect if a file has a `.sia` extension and upload it properly if so.

app/plugins/Files/js/loaders.js
pmknutsen: @lukechampine `fs.existsSync` is (or: is being) deprecated.
I dropped the existence check altogether

app/plugins/Files/js/buttons.js
pmknutsen, line 167: @Mingling94 I don't think its necessary. If the path disappears, or is invalid, this is handled in `uploadDir()` in `loaders.js`. It seems redundant to check for valid path twice. Let me know if you had something else in mind.

app/plugins/Files/js/buttons.js
pmknutsen, line 174: @Mingling94 Both forward and backward slash is replaced with underscore.

app/plugins/Files/index.html
pmknutsen: @Mingling94 Let me know if you like the `stepf1` scheme better.

app/plugins/Files/js/buttons.js
pmknutsen: @Mingling94 Moved requires out to `helpers.js`

app/plugins/Files/js/buttons.js
Mingling94, line 167: It can be a todo. Currently, one could be using an invalid loadpath and not enter this if-statement, confusing a somewhat unaware user

app/plugins/Files/js/buttons.js
Mingling94, line 174: Okay cool, so the file gets a nick name involving the directory it's in? So a file of `C:\\Users\Ming\Documents\test.txt` becomes `C:__Users_Ming_Documents_test.txt`?

Mingling94:
If this passes some initial testing/screwing-around, I'm all for merging it. I've not got the time to test it though

app/plugins/Files/js/buttons.js
pmknutsen, line 174: The gets a prefix, which is the folder containing the files: `C:\\Users\Ming\Documents\test.txt` becomes `Documents_test.txt` by default. The user can edit the prefix (ie `Documents_`) before uploading. The reason for the substitution was errors when `/` was included in nicknames. For future reference, it would be nice to allow `/` and `\` in nicknames as one could then display a virtual file structure in the Sia-UI.

DavidVorick:
We will be testing it today and tomorrow, if all goes well we'll do another pass over the code and then merge it.

Sorry we've been pretty inactive with PRs the past 2 weeks. Been quite busy.

app/plugins/Files/js/loaders.js
```diff
@@ -87,6 +112,14 @@ addResultListener('ascii-loaded', function(result) {
 	update();
 });
 
+//Confirm deletion popup
+function confirmDelete(nickname) {
+	eID('confirm-delete').querySelector('.title').innerHTML = 'Confirm to delete';
```
lukechampine: any reason why this isn't just defined in the HTML?
I would change the title to `'Confirm'`, and inside the popup `'Really delete ' + nickname + '?'`

also, in the future, try not to add multiple features in one pull request.

app/plugins/Files/js/buttons.js
lukechampine, line 174: > display a virtual file structure in the Sia-UI

this is indeed a planned feature, though we haven't settled on the right way to do it yet. :)

lukechampine:
I tested this today and it worked as expected. Just need to tweak the text of the delete confirmation dialog and I'll merge it.

app/plugins/Files/js/loaders.js
pmknutsen: You're right: I think the confirmation string could just be defined in HTML.

And yes, I realize there were too many changes in both of my pull requests. I was oblivious to the constraints of a multi-developer effort. Future PRs will be much smaller and more focused :-)

app/plugins/Files/js/buttons.js
pmknutsen, line 174: I think such a feature needs front-seat as for setting UI priorities. The current linear browser will get clunky fast with a large number of file uploads. A straight-forward hack is to denote directories in the nick, e.g. `%%secrets%%lionelrichie%%blackmail.pdf` becomes `/secrets/lionelrichie/blackmail.pdf`. A directory browser would be very easy to implement this way.

DavidVorick:
_poke_ @pmknutsen I'm eager to merge this, what's the status on the delete dialog?

pmknutsen:
@DavidVorick Done. The nick to be deleted is fetched form the innerHTML of `.nickname`, so I left the string inside the dialog as-is for now, rather than making more changes elsewhere.

DavidVorick:
You can't close the PR, it has to be open for us to merge it.

PR #80, opened 2015-10-06
CSS restyling and Wallet improvements

pmknutsen:
This PR introduces new CSS styling and added functionality in the Wallet plugin, such as scrollable address and transaction lists. Users can view transactions of wallet addresses by selecting an address in the address list. The address list can be filtered to quickly bring up transactions of interests, such as deposits. The Wallet header is now minimized when the wallet is locked.

The PR also updates the Files plugin, with icons replacing text links and CSS improvements to the list layout. The text in the Hosting plugin has been replaced with more descriptive strings.

siaui_css_restyling_address_search


app/plugins/Wallet/css/wallet.css
```diff
@@ -126,3 +128,99 @@ select:hover {
...
+	overflow: hidden;
+}
+
+/* Transactions list */
+#address-list {
+	background-color: #fafafa;
+	height: calc(100% - 35px);
+	width: 100%;
+	padding: 0px;
+	overflow-y: scroll;
```
Mingling94: Should be `overflow-y: auto;` to hide the scroll bar when it's unneeded.

app/plugins/Wallet/js/loaders.js
```diff
@@ -33,7 +32,7 @@ addResultListener('locked', function(result) {
 // Markup changes to reflect unlocked state
 function setUnlocked() {
 	clearLockIcon();
-	eID('lock-status').innerHTML = 'Unlocked';
+	eID('lock-status').innerHTML = 'Lock Wallet';
```
Mingling94, line 35: Love it, clicking an action and not a status!

app/plugins/Wallet/js/lifecycle.js
```diff
@@ -1,89 +1,163 @@
...
+	if (typeof(txn) == 'undefined') { return; }
+
+	// Add only new transactions
+	if (eID(txn.transactionid)) { return; }
+	var txnElement = eID('transactionbp').cloneNode(true);
+	txnElement.id = txn.transactionid;
+	txnElement.timestamp = txn.confirmationtimestamp * 1000;
+
+	// DOM shortcut
+	var field = function(selector) {
```
Mingling94, line 103: Oh this was another possible lag-point. I'm not sure if defining this function for temporary usage in `appendTransaction` opens up the possibility of a memleak or something.

app/plugins/Wallet/js/lifecycle.js
```diff
@@ -1,89 +1,163 @@
...
+	if (txn.outputs) {
+		txn.outputs.forEach( function(output) {
+			if (output.walletaddress) {
+				amount = amount.add(output.value);
+			}
+		});
 	}
-	var entry = eID('transactionbp').cloneNode(true);
-	entry.id = txn.transactionid;
+	var ConversionFactor = new BigNumber(10).pow(24);
```
Mingling94: There's a helper.js function for this.

app/plugins/Wallet/js/lifecycle.js
```diff
@@ -1,89 +1,163 @@
...
-	// Have to use !== logic to represent miner payouts
-	var sign, unit, amount, related;
-	var ft = txn.fundtype.split(' ');
-	sign = ft[1] !== 'input' ? '+' : '-';
-	amount = ft[0] !== 'siafund' ? convertSiacoin(txn.value) : txn.value;
-	unit = ft[0] !== 'siafund' ? ' Siacoin ' : ' Siafund ';
-	related = ft[1] !== 'input' ? ' received into ' : ' sent from ';
-	related += txn.relatedaddress;
+	// Format transaction timestamp
+	var timestamp = new Date(txn.confirmationtimestamp * 1000);
```
Mingling94, line 130: You do this timestamp \* 1000 multiplication twice. Perhaps make a helper function for it?

app/plugins/Wallet/js/lifecycle.js
```diff
@@ -1,89 +1,163 @@
...
+
+	/* Fetch all wallet transactions by iterate over wallet addresses
+	var loopmax = result.addresses.length;
+	var counter = 0;
+	(function next() {
+		setTimeout(function() {
+			updateAddrTxn(result.addresses[counter].address);
+			next();
+		}, 50); // force 50 ms delay between each GET request
+	})();*/
```
Mingling94, line 37: What is this commented out stuff? A defined function that then gets executed? Why?

app/plugins/Wallet/js/buttons.js
```diff
@@ -11,6 +11,22 @@ eID('create-address').onclick = function() {
...
 
+// Filt address list by search string
+eID('search-address').onclick = function() {
+	tooltip('Searching...', this);
+	var searchstr = eID('search-bar').value;
+
+	NodeList.prototype.forEach = Array.prototype.forEach
+	var entries = eID('address-list').childNodes;
+	entries.forEach( function(entry) {
+		if (entry.getElementsByClassName('address')[0].innerHTML.indexOf(searchstr)) {
```
Mingling94: `querySelector()` will get you the first element you need, negating the need to have `[0]`

app/plugins/Wallet/index.html
```diff
@@ -6,11 +6,24 @@
...
+			<div class='type'>
+				<div class='send'>
+					<i class='fa fa-sign-out'></i>
+				</div>
+				<div class='receive'>
+					<i class='fa fa-sign-in'></i>
+				</div>
 			</div>
+			<div class='amount'></div>
+			<div class='tnxid'></div>
```
Mingling94: Should be txnid?

app/plugins/Wallet/css/wallet.css
```diff
@@ -115,9 +117,9 @@ select:hover {
 	height: calc(50% - 46px);
 }
 .list {
-	height: calc(100% - 108px);
+	height: calc(50%);
 	padding: 16px;
-	overflow:auto;
+	overflow:hide;
```
Mingling94: Should keep the space after the colon in CSS since that's been the convention.

We could have a CSS linter fire for this.

app/plugins/Wallet/css/wallet.css
```diff
@@ -79,35 +88,28 @@
...
-	font-family: sans-serif;
-	font-size: 18px;
-	box-shadow: none;
-	color:#4a4a4a;
-
-	border: solid 1px #dddddd;
-	transition: box-shadow 0.3s, border 0.3s;
+#transaction-amount {
+	width: 100px;
+	text-align:center;
```
Mingling94: ditto

app/plugins/Wallet/css/wallet.css
```diff
@@ -79,35 +88,28 @@
 .transparent {
 	-webkit-animation: fadeout 1s; /* Safari, Chrome and Opera > 12.1 */
 	opacity: 0;
+	display:none;
```
Mingling94: and a bunch of other places. Regex for it.

Mingling94:
Awesome look! I just read code and commented stuff. Yet to test (perhaps a testing framework is needed down the road)

app/plugins/Wallet/js/lifecycle.js
pmknutsen, line 37: So, this was a temporary fix to be able to actually view the entire transaction history without the UI freezing. The loop retrieves all transactions _per_ address, and then pauses 50 ms before retrieving transactions of the next address (edit: the pause was necessary or the UI went haywire). This way, many small datasets are returned and the transaction list in the UI updates in steps. The drawback is that its a real CPU drain and takes a while to execute when you have a large number of addresses. I thus opted for the view-transactions-when-clicking-an-address solution, but left this code in there perhaps as a future option if a user wants to view _all_ transactions.

app/plugins/Wallet/css/wallet.css
pmknutsen: Fixed in 3e9ada4139a137c051bb5d105262f70820fc782d

app/plugins/Wallet/js/lifecycle.js
pmknutsen, line 103: I don't quite see an issue. Should it not be destroyed when leaving `appendAddress`? In any case, I have copied over the TODO memo from `Files/lifecycle.js`. 5151970149d734c62b9e0e638c47cd60f7322363

app/plugins/Wallet/js/lifecycle.js
pmknutsen: Fixed in 457c3a508d413e9b33dec8d6ff4d80c22eb79617

app/plugins/Wallet/js/lifecycle.js
pmknutsen, line 130: Hmm, a helper function for a single multiplication seems unnecessary.

Mingling94, line 130: Word, it should be fine

app/plugins/Wallet/js/buttons.js
pmknutsen: Fixed in ef965b91fb879335726c39ce266c8087c8b14578

app/plugins/Wallet/index.html
pmknutsen: Tnx for noticing ;-) Fixed in f965241a40d80d1bd1e0eaaa667c5ac4169136e5

DavidVorick:
testing framework would be good, as soon as we feel comfortable that we've hit feature-minimums

app/plugins/Wallet/css/wallet.css
pmknutsen: Fixed in 48b7ccfdfbb875bd0052a4eda41e8140eb7f0f09

app/plugins/Wallet/css/wallet.css
pmknutsen: Fixed in 48b7ccfdfbb875bd0052a4eda41e8140eb7f0f09

app/plugins/Wallet/css/wallet.css
pmknutsen: Fixed all over, in 48b7ccfdfbb875bd0052a4eda41e8140eb7f0f09

app/plugins/Wallet/js/loaders.js
```diff
@@ -74,11 +73,11 @@ IPC.on('unlocked', function(err, result) {
 });
 
 // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ Encrypting ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-// setUnencrypted sets the wallet lock status to unencrypted.
+// setUnencrypted sets the wallet lock status to encrypted.
```
lukechampine: something has to be changed here because this docstring makes no sense

app/plugins/Wallet/js/loaders.js
```diff
@@ -74,11 +73,11 @@ IPC.on('unlocked', function(err, result) {
...
 function setUnencrypted() {
 	clearLockIcon();
-	eID('lock-status').innerHTML = 'Create Wallet';
-	eID('lock-icon').classList.add('fa-plus');
+	eID('lock-status').innerHTML = 'Unencrypted';
```
lukechampine: Why change this from an action to a status?

lukechampine:
Looks very slick! General comments:
- Transactions should be sorted newest-oldest.
- Searching addresses should happen as you type. The search button can be removed entirely. Also, a nice feature would be to highlight and/or scroll to the address that is created by the "Create Address" button.
- Tray icon should be updated to the new logo.
- Long file sizes (e.g. `404 Bytes`) are wrapping around on my display, such that `Bytes` is on a newline. There's quite a bit of room to expand this field leftward.
- Loading transactions is still quite slow, causing 2-3 second hangs. This is definitely something we need to address. Doesn't need to be in this PR though.

pmknutsen:
@lukechampine I'll commit the requested changes with an ETA of Oct 23rd due to other commitments. As for slow loading when displaying all transactions: I think the `wallet/transactions/` API call needs to change so that pagination is possible, e.g. an optional parameter `count` to return N last transactions of the retrieved dataset.

lukechampine:
ok. We are moving fairly quickly at the moment, so we may opt to merge this sooner rather than later and make the changes ourselves, since they are relatively minor.
I will look into the slow loading issue myself and see if I can uncover anything. In the meantime, we should at least make the calls asynchronously so that switching between views is instant.

app/plugins/Wallet/css/wallet.css
Mingling94: What about this particular one, then?

app/plugins/Wallet/js/loaders.js
Mingling94: Yeah this comment should be reverted. `setUnencrypted` does what it says.

app/plugins/Wallet/js/loaders.js
Mingling94: Agreed, should also revert this part.

Mingling94:
@lukechampine I think this PR should be merged after the changes I just commented on (as they're <5 minute tasks). Making the calls asynchronous could perhaps draw out the PR quite a bit longer.

DavidVorick:
@Mingling94 @pmknutsen how hard would it be to make the transactions a separate menu within the wallet, such that the data is not loaded until the menu is accessed?

I think it's very important that the wallet tab load immediately. It's okay if that means that the transactions don't get loaded immediately and instead the user needs to click on a 'show transactions' button which then takes 1-10 seconds. It's less disruptive to the overall experience that way.

app/plugins/Wallet/css/wallet.css
pmknutsen: Fixed in de087eba964bc76065193adc08c8a0ba2a42ec62

app/plugins/Wallet/js/loaders.js
pmknutsen: Reverted `setUnencrypted` to original in f4c4c08ca1012ec5fb17baeb1d917adeb65c2620

Mingling94:
@DavidVorick It wouldn't be hard to tie loading the transactions to a button above the window. Since this is @pmknutsen's PR though, it depends on when he can get to it.

DavidVorick:
The transactions thing was actually merged into master in another PR, so it shouldn't hold this PR back

pmknutsen:
@DavidVorick My previous PR was for uploading dirs, in Files plugin, not transactions.

Working on some of the requests by @lukechampine now and will have them ready shortly.

pmknutsen:
@lukechampine @Mingling94 @DavidVorick Commit 5e8d426d21d9d66ba4f39461a4b2cfdbaedc4e8a adds a refresh button to load all wallet transactions, rendering the wallet faster. There is still some slowness if the wallet has a large number of addresses. My wallet for some reason has >9000 addresses in it, which takes 2-3 sec to load and render.

app/plugins/Wallet/js/buttons.js
```diff
@@ -11,6 +11,22 @@ eID('create-address').onclick = function() {
 	IPC.sendToHost('api-call', call, 'new-address');
 };
 
+// Filter address list by search string
+eID('search-bar').onkeydown = function() {
```
lukechampine: need to use `onkeyup` here; otherwise `searchstr` will contain the text in the search bar _prior_ to the keypress.

app/plugins/Wallet/js/buttons.js
```diff
@@ -25,8 +41,8 @@ function appendAddress(address) {
 
 // Add the new address
 addResultListener('new-address', function(result) {
-	notify('Address created!', 'created');
-	appendAddress(result.address);
+	notify('New address created', 'created');
+	appendAddress(result);
```
lukechampine, line 45: this will add the new address even if it doesn't match the current search term. Adding addresses is a bit janky now anyway, so maybe I'll just revisit this later.

lukechampine:
ok, looking good to me. I'm ready to merge this unless you want to add more within the next 24 hours

pmknutsen:
I'm done with changes for now.

lukechampine:
again -- really appreciate your work here. The new styling looks great!

PR #82, opened 2015-10-07
Prevent Leaking arguments

bitspill:
Remove arugment leaks, https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#32-leaking-arguments

This also speeds up `/wallet/transactions` IPC calls as encountered by @pmknutsen on Slack

### DIFF
--- a/app/js/plugin.js
+++ b/app/js/plugin.js
@@ -77,7 +77,12 @@ function Plugin(plugPath, name) {
 		* @param {...*} messages - ipc messages sent over channel to view
 		*/
 		sendToView: function() {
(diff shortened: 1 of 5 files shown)


Mingling94:
Oh cool! Are there noticeable improvements?

pmknutsen:
Yes, this seems to be an improvement. My entire transaction history (~1650 transactions, totaling 3.2 MB of JSON data) is now displayed in about 7 sec (plus ~1sec for retrieving the data).

lukechampine:
awesome. I suspect that `sendToView` was the biggest culprit, but it doesn't hurt to fix the others as well.

PR #83, opened 2015-10-08
Local copy of Roboto-Condensed for offline use.

bitspill:


DavidVorick:
Roboto-Condensed was already local? I'm not sure what this achieves.

bitspill:
Inside `app/css/roboto-condensed-min.css` this was a ToDo because currently it is pulled from Google Fonts
`/* TODO: Get roboto http imports so that this works when offline */`
and
`url(http://fonts.gstatic.com/s/robotocondensed/v13/b9QBgL0iMZfDSpmcXcE8nE8xhg-WlD4iG4AP-V7QxYA.woff2) format('woff2')`

This commit replaces the references to Google Fonts and refers to local .ttf files

Mingling94:
Yasssss! I totally forgot I made that TODO! I noticed roboto not loading when I ran Sia-UI while coding on a plane without internet.

doc/PluginTutorial.md
```diff
@@ -164,7 +164,7 @@ section:
 ```html
 	<head>
 		<!-- CSS -->
-		<link rel="stylesheet" href="../../css/roboto-condensed-min.css">
+		<link rel="stylesheet" href="../../css/roboto-condensed/css/roboto-condensed-min.css">
```
Mingling94: Uhhh, can we make this folder path not so redundant?

bitspill: I almost left the .css file where it was only adding the fonts folder but I ended up modeling it after font-awesome's directory structure, if preferred I can move the .css back to a simpler file path

DavidVorick: I agree with Ming, try to keep the directory structure as simple as possible.

Some of the node.js stuff is in love with layered modules and whatnot, but it can make it difficult to find things and often causes problems in Windows where filepaths are only allowed to be 255 characters (starting from C:) in the first place.

Mingling94: The font-awesome folder structure was a lazy move. Let's make the css folder one layer deep ^_^

bitspill:
Simplified the directory structure.

lukechampine:
@Mingling94, you have the go-ahead to merge this if it looks good.

PR #86, opened 2015-10-12
Periodically check wallet plugin status

bitspill:
If you launch the UI and quickly tab over to Wallet before the daemon is launched and checked for updates the tab will be in a "stuck" state without updating it's status.

The initial update check is aborted `if (!Daemon.Running)` which isn't set until a few seconds after launch, so periodically try again until the daemon is online and responds.

### DIFF
--- a/app/plugins/Wallet/js/lifecycle.js
+++ b/app/plugins/Wallet/js/lifecycle.js
@@ -2,6 +2,10 @@
 
 var blockheight = 0;
 
+// How often /wallet updates
+var refreshRate = 500; // half-second
+var finalRefreshRate = 1000 * 60 * 5; // five-minutes
+
 // Library for working with clipboard
 const Clipboard = require('clipboard');
 
@@ -16,8 +20,7 @@ function update(address) {
 		type: 'GET',
 	}, 'update-address');
 
-	updating = setTimeout(update, 60000 * 5); // update every 5 min
-	return;
+	updating = setTimeout(update, refreshRate);
 }
 
 // Add transactions to view list per address
@@ -51,6 +54,18 @@ function updateAddrTxn(addr) {
 
 // Update wallet summary in header
 addResultListener('update-status', function(wallet) {
+	refreshRate = finalRefreshRate; // slow down after first successful call
+
+	// Show correct lock status.
+	// TODO: If the wallet is encrypted, prompt with a pw.
+	if (!wallet.encrypted) {
+		setUnencrypted();
+	} else if (!wallet.unlocked) {
+		setLocked();
+	} else if (wallet.unlocked) {
+		setUnlocked();
+	}
+
 	// Update balance confirmed and uncomfirmed
 	var bal = convertSiacoin(wallet.confirmedsiacoinbalance);
 	var pend = convertSiacoin(wallet.unconfirmedincomingsiacoins).sub(convertSiacoin(wallet.unconfirmedoutgoingsiacoins));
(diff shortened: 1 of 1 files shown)


app/plugins/Wallet/js/lifecycle.js
```diff
@@ -91,14 +91,17 @@ function start() {
 	// DEVTOOL: uncomment to bring up devtools on plugin view
 	// IPC.sendToHost('devtools');
 	
-	// Need to check if wallet's unencrypted
-	IPC.sendToHost('api-call', '/wallet', 'on-opened');
+	// Need to check if wallet's unencrypted, keep checking until we get a response
+	updating = setInterval(IPC.sendToHost, 500, 'api-call', '/wallet', 'on-opened');
```
Mingling94: I would do this in a slightly different way. You're using the same variable, updating, that the update function uses and that's confusing its purpose. Make a new variable called `starting`?

bitspill: I considered that however to me both of these uses were "updating" so it seemed to fit.

I'll move it over to a seperate `starting` variable for more clarity.

Mingling94: Actually, it makes more sense to just combine the 'on-opened' and 'update-status' listeners. They're the same call and data, so why not

bitspill: Yea, that does sound reasonable, done.

bitspill:
Updated after the wallet restyling merge.

app/plugins/Wallet/js/lifecycle.js
```diff
@@ -46,8 +48,20 @@ function updateAddrTxn(addr) {
 	}, 'update-history');
 }
 
-// Update wallet summary in header
+// Update transaction history and addresses
 addResultListener('update-status', function(wallet) {
+	refreshRate = finalRefreshRate; // slow down after first successful call
+
+	// Show correct lock status.
+	// TODO: If the wallet is encrypted, prompt with a pw.
```
DavidVorick, line 60: password prompt should not appear until the user hits the 'encrypt' button. Standard instinct when dealing with app popups is to TL;DR and close them, then people don't know their password.

bitspill, line 60: That's an old ToDo, I adjusted the line formatting and moved it to a different location so git shows it as added. 

https://github.com/NebulousLabs/Sia-UI/blob/master/app/plugins/Wallet/js/lifecycle.js#L183

Mingling94, line 60: What I meant when I wrote the TODO is that when opening the wallet plugin, the user should be prompted to enter their password. Unless it's viable that the plugin is used meaningfully when locked, the user should login on opening, right?

DavidVorick, line 60: ok, I think this TODO can be removed at this point

DavidVorick, line 60: @Mingling94 I think we commented at the same time.

After you unlock the wallet the first time, it's viable to see your balance and txn history while the wallet is locked. I'm planning some upgrades so that this is also true even if you haven't yet unlocked the wallet. Also, it will eventually be okay to mine + do some forms of hosting and renting while the wallet is locked. None of these features are yet available, but they are in the pipeline.

bitspill:
Re-based after recent merges.

PR #87, opened 2015-10-12
Ignore IntelliJ project files.

bitspill:
For those that can't vim


DavidVorick:
I strongly recommend learning vim. https://danielmiessler.com/study/vim/

And then reading tips+tricks every few weeks to expand your vim vocabulary. Once you are comfortable with it, it completely blows away anything you can do with a mouse.

PR #88, opened 2015-10-13
Don't show 0 balance when not Unlocked

bitspill:
By showing hyphens rather than a 0 it should put some minds at ease rather than assuming all coins are lost.

### DIFF
--- a/app/plugins/Overview/js/overview.js
+++ b/app/plugins/Overview/js/overview.js
@@ -69,11 +69,12 @@ function tooltip(message, element) {
 
 // Define IPC listeners and update DOM per call
 IPC.on('wallet-update', function(err, result) {
-	var bal = result !== null ? formatSiacoin(result.confirmedsiacoinbalance) : null;
-	updateField(err, 'Balance: ', bal, 'balance');
+	if(!result){
+		return;
+	}
 
-	var unlocked = result !== null ? result.unlocked : null;
-	var encrypted = result !== null ? result.encrypted : null;
+	var unlocked = result.unlocked;
+	var encrypted = result.encrypted;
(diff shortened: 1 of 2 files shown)


DavidVorick:
Cool. We should probably do this in siac as well

app/plugins/Overview/js/overview.js
```diff
@@ -81,6 +80,8 @@ IPC.on('wallet-update', function(err, result) {
 	} else {
 		updateField(err, 'Locked', '', 'lock');
 	}
+
+	updateField(err, 'Balance: ', unlocked ? bal : '---', 'balance');
```
Mingling94, line 87: This should go up near where `bal` is initialized. Further, the whole function does nothing when `result === null` so a 

```
if (!result) {
    return;
}
```

at the top would make the code less verbose and more readable.

lukechampine:
Didn't https://github.com/NebulousLabs/Sia-UI/commit/75955d0920c7e33c38bad7f8c095ad3333bb7a67 address this? I don't see a balance at all when the wallet is locked.

bitspill:
https://github.com/NebulousLabs/Sia-UI/commit/75955d0920c7e33c38bad7f8c095ad3333bb7a67 only modified the Wallet plugin, this pull is for the Overview plugin.

app/plugins/Overview/js/overview.js
bitspill, line 87: The balance field is dependent upon the unlock state so it had to be moved below it, I just brought the `bal` initialization down next to it so they're together.

`result === null` adjustment has also been made.

lukechampine:
ah, gotcha

PR #90, opened 2015-10-15
Add tooltips to file action icons

bitspill:
The icons may be slightly ambiguous, this adds simple tooltips on mouse hover

### DIFF
--- a/app/plugins/Files/css/files.css
+++ b/app/plugins/Files/css/files.css
@@ -279,3 +279,24 @@ input:focus {
 	color: #777777;
 	cursor: pointer;
 }
+
+.simpleTooltip {
+	display: inline;
+	position: relative;
+}
+
+.simpleTooltip:hover:after {
+	background: #111;
+	background: rgba(0,0,0,.8);
+	border-radius: .5em;
+	top: 1.35em;
+	color: #fff;
+	content: attr(data-tooltip);
+	display: block;
+	right: 1em;
+	padding: .3em 1em;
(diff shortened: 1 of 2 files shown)


DavidVorick:
everybody's price estimation is all over the place. I guess I should take a second look at the way that is assembled.

edit: mine is currently 13, but I've seen ranges between 13 and 2000 over the past few days. Maybe it's because there's currently a high host churn.

app/plugins/Files/css/files.css
```diff
@@ -279,3 +279,24 @@ input:focus {
...
+.simpleTooltip:hover:after {
+	background: #111;
+	background: rgba(0,0,0,.8);
+	border-radius: .5em;
+	top: 1.35em;
+	color: #fff;
+	content: attr(data-tooltip);
+	display: block;
+	right: 1em;
+	padding: .3em 1em;
```
DavidVorick, line 297: is there a standard that we are using? I think that most of the UI is currently in `px`

Mingling94:
I wasn't aware there was a simple `data-tooltip` attribute you could add for div elements. Can this replace the event-based tooltips that are communicated over IPC?

bitspill:
There is nothing inherently special about data-tooltip it's just a generic data-\* attribute with corresponding CSS. 
https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_data_attributes

I'll see if I can work it into replace the IPC tooltips.

PR #93, opened 2015-10-17
Fix issue with a bad rebase/merge on TooSoon.

bitspill:
When rebasing the TooSoon branch just before the merge I merged this section incorrectly.

### DIFF
--- a/app/plugins/Wallet/js/lifecycle.js
+++ b/app/plugins/Wallet/js/lifecycle.js
@@ -197,28 +197,9 @@ addResultListener('update-history', function(result) {
 function start() {
 	// DEVTOOL: uncomment to bring up devtools on plugin view
 	// IPC.sendToHost('devtools');
-	
-	// Need to check if wallet's unencrypted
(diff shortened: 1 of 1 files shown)


app/plugins/Wallet/js/lifecycle.js
```diff
@@ -197,28 +197,9 @@ addResultListener('update-history', function(result) {
...
-   	} else if (!wallet.unlocked) {
-		setLocked();
-	} else if (wallet.unlocked) {
-		setUnlocked();
-	}
-
-	// Start updating
 	update();
-});
+}
```
DavidVorick, line 202: suspicious...

But looks like it'll work. I'm just remembering we had problems the last time this diff was introduced. But here it's closing a standard function instead of a closure.

bitspill, line 202: Heh, yea this should be the correct form. Last time I accepted the rebase too hastily and did not verify it manually. On startup `update ()` should be called directly and the `on-opened` IPC call should no longer exist.

PR #96, opened 2015-10-20
Better daemon update checking

Mingling94:
Also re-added the siad-import npm commands in the package.json

### DIFF
--- a/app/js/daemonManager.js
+++ b/app/js/daemonManager.js
@@ -72,17 +72,21 @@ function DaemonManager() {
 		// duration that it takes to load up the blockchain.
 		apiCall("/daemon/updates/check", function(err, update) {
 			if (err) {
+				console.log(err, update)
 				self.Running = false;
 				// Check again later
-				setTimeout(updatePrompt, 1000);
+				setTimeout(updatePrompt, 5000);
 			} else if (update.Available) {
 				self.Running = true;
 				UI.notify("New Sia Client Available: Click to update to " + update.Version, "update", function() {
 					Shell.openExternal('https://www.github.com/NebulousLabs/Sia-UI/releases');
 				});
-			} else {
+			} else if (!update.Available){
 				self.Running = true;
 				UI.notify("Sia client up to date!", "success");
+			} else {
+				console.log("Unknown result from update check!");
+				console.error(err, update);
 			}
 		});
 	}
@@ -96,19 +100,13 @@ function DaemonManager() {
 		}, function() {
 			UI.notify('Starting siad...', 'start');
 
-			// daemon as a background process logs output to files
-			var out = Fs.openSync(Path.join(__dirname, siaPath, 'daemonOut.log'), 'w');
-			var err = Fs.openSync(Path.join(__dirname, siaPath, 'daemonErr.log'), 'w');
-
 			// daemon process has to be detached without parent stdio pipes
 			var processOptions = {
 				detached: false,
-				stdio: [ 'ignore', out, err ],
 				cwd: Path.join(__dirname, siaPath),
 			};
 			var command = process.platform === 'win32' ? './siad.exe' : './siad';
 			var daemonProcess = new Process(command, processOptions);
-			daemonProcess.unref();
 
 			// Give siad time to load or exit
 			var updating = setTimeout(updatePrompt, 1000);
--- a/package.json
+++ b/package.json
@@ -18,6 +18,10 @@
     "debug": "node-inspector & electron --debug=5858 . & xdg-open http://127.0.0.1:8080/debug?ws=127.0.0.1:8080&port=5858",
     "doc": "jsdoc -c .jsdocrc",
     "lint": "jshint . --verbose --exclude-path=.jshintignore",
+    "sia-repo": "go get -u github.com/NebulousLabs/Sia/...",
+    "sia-import": "(test -d app/Sia || mkdir -p app/Sia) && cp $GOPATH/bin/siad app/Sia",
+    "sia-kill": "killall siad; while pgrep siad; do sleep .5; done;",
+    "sia-reimport": "npm run sia-repo && npm run sia-kill && npm run sia-import",
     "release": "./release.sh"
   },
   "repository": {
--- a/.gitignore
+++ b/.gitignore
@@ -2,8 +2,9 @@
 release
 node_modules
 npm-debug.log
-app/config.json
-app/dependencies
+config.json
+dependencies
+Sia
 doc/Sia-UI
 
 # Vim


app/js/daemonManager.js
```diff
@@ -72,17 +72,21 @@ function DaemonManager() {
 		// duration that it takes to load up the blockchain.
 		apiCall("/daemon/updates/check", function(err, update) {
 			if (err) {
+				console.log(err, update)
 				self.Running = false;
 				// Check again later
-				setTimeout(updatePrompt, 1000);
+				setTimeout(updatePrompt, 5000);
```
DavidVorick, line 78: why 5 seconds?

app/js/daemonManager.js
```diff
@@ -72,17 +72,21 @@ function DaemonManager() {
...
 				self.Running = true;
 				UI.notify("New Sia Client Available: Click to update to " + update.Version, "update", function() {
 					Shell.openExternal('https://www.github.com/NebulousLabs/Sia-UI/releases');
 				});
-			} else {
+			} else if (!update.Available){
 				self.Running = true;
 				UI.notify("Sia client up to date!", "success");
+			} else {
+				console.log("Unkown result from update check!");
```
DavidVorick: typo

package.json
```diff
@@ -18,6 +18,11 @@
     "debug": "node-inspector & electron --debug=5858 . & xdg-open http://127.0.0.1:8080/debug?ws=127.0.0.1:8080&port=5858",
     "doc": "jsdoc -c .jsdocrc",
     "lint": "jshint . --verbose --exclude-path=.jshintignore",
+    "postinstall": "npm run sia-repo && npm run sia-import",
```
DavidVorick: this returns errors if you got the Sia repo through ssh

package.json
```diff
@@ -18,6 +18,11 @@
...
+    "sia-repo": "go get -u github.com/NebulousLabs/Sia/...",
```
DavidVorick, line 21: specifically, this returns errors if you got 'siad' through ssh

app/js/daemonManager.js
```diff
@@ -96,19 +100,13 @@ function DaemonManager() {
 		}, function() {
 			UI.notify('Starting siad...', 'start');
 
-			// daemon as a background process logs output to files
-			var out = Fs.openSync(Path.join(__dirname, siaPath, 'daemonOut.log'), 'w');
-			var err = Fs.openSync(Path.join(__dirname, siaPath, 'daemonErr.log'), 'w');
```
DavidVorick, line 101: is there somewhere else that the files go? I feel like these files would be really useful to have, though I forgot about them and haven't asked users for them yet.

package.json
```diff
@@ -18,6 +18,11 @@
...
+    "sia-import": "(test -d app/dependencies/Sia || mkdir -p app/dependencies/Sia) && cp $GOPATH/bin/siad app/dependencies/Sia",
+    "sia-kill": "killall siad; while pgrep siad; do sleep .5; done;",
```
DavidVorick, line 23: I'm not sure I understand why this rule is needed

app/js/daemonManager.js
Mingling94, line 78: 1 second is too frequent to easily read debug output while 5 seconds isn't really that noticeable of a wait imo

package.json
```diff
@@ -18,6 +18,11 @@
     "debug": "node-inspector & electron --debug=5858 . & xdg-open http://127.0.0.1:8080/debug?ws=127.0.0.1:8080&port=5858",
     "doc": "jsdoc -c .jsdocrc",
     "lint": "jshint . --verbose --exclude-path=.jshintignore",
+    "postinstall": "npm run sia-repo && npm run sia-import",
+    "sia-repo": "go get -u github.com/NebulousLabs/Sia/...",
+    "sia-import": "(test -d app/dependencies/Sia || mkdir -p app/dependencies/Sia) && cp $GOPATH/bin/siad app/dependencies/Sia",
```
DavidVorick: I don't know why we need this rule - it hasn't been needed for installation for Luke + I. But maybe that's because we haven't done any clean installs since the rule got removed.

package.json
Mingling94, line 21: I see, I'll remove the postinstall line but the other commands do no harm to stay

app/js/daemonManager.js
Mingling94, line 101: I got an error once from these lines. Since siad is no longer a background process, these messages go to their parent process's console, e.g. the general UI's

app/js/daemonManager.js
DavidVorick, line 78: didn't realize it's an update check.

5 seconds is probably more than needed too. Even 10 minutes should be fine as far as software updates go.

Mingling94, line 78: 5 seconds is the time to retry updating if it fails. While this hasn't succeeded, Daemon.Running is false and no api calls are being made.

package.json
Mingling94, line 23: To simulate siad not working/going offline for debugging purposes

package.json
Mingling94: How do you update the siad in the UI?

app/js/daemonManager.js
bitspill, line 78: Unrelated to the pull at hand, is there a way to check Daemon.Running from a plugin? I couldn't find it.

Mingling94, line 78: If you look at pluginManager.js, I implemented a trick where:
if a plugin's apiCall fails
Daemon.ifSiad() is triggered
    which will flip Daemon.Running if that fails too
    or will send back the error and null result if ifSiad() shows siad is running

You could add another ipc channel specifically for this, though

DavidVorick, line 78: I definitely don't think that you should be flipping `Daemon.Running` based on the outcome from a call to `/daemon/updates/check`. The only api call that should be really used to check if the daemon is working is `/daemon/version`

Mingling94, line 78: Yep, `ifSiad()` does precisely that, turning it false if daemon hasn't started. Then it starts it, waiting until an update check returns true to flip Daemon.Running to true.

I _suppose_ since Luke made the update check pointless that updatePrompt() should just be turned into

```
function checkSiad() {
    ifSiad(function() {
        UI.notify('siad started!', 'check');
    }, function() {
        UI.notify('Waiting for siad', 'ongoing');
        // Check again later
        setTimeout(checkSiad, 1000);
    });
}
```

Mingling94, line 78: @DavidVorick Thoughts on this approach? It would be an entirely different PR

PR #97, opened 2015-10-20
Fix problematic config.json

pmknutsen:
Revert to default configuration if config.json is `undefined`


lukechampine:
LGTM. Do we know why the config.json would be undefined, though?

DavidVorick:
we don't, which is why this is an unfinished issue. But having this temp fix is better than having no fix.

PR #100, opened 2015-10-22
wait for siad to start

lukechampine:
`self.Running` is now properly set, and a notification will now be displayed when siad starts responding.

We really ought to display an uncloseable "Loading" dialogue until siad starts responding. The `waitForSiad` function should make this fairly easy to implement, it's just a matter of managing the popup html properly.


Mingling94:
:+1:

PR #103, opened 2015-10-23
Some changes that I think are needed

Mingling94:
Perhaps I don't understand .gitignore files, but I don't understand how
the hell your app/Sia folder isn't being included with commits and shiz.

Also there's no longer such thing as a 'dependencies' folder.

Also also the sia import scripts were useful, please re-add them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,8 +2,8 @@
 release
 node_modules
 npm-debug.log
-app/config.json
-app/dependencies
+config.json
+Sia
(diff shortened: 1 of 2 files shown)


DavidVorick:
probably because we don't ever import Sia, we might just use the UI with the daemon that's already running.

DavidVorick:
That's probably also why we killed the 'import' command.

PR #105, opened 2015-10-26
Small usability fixes

### LINKED ISSUE #101: Window size not preserved
Currently the main window size is not preserved when the client gets shut down and restarted. Instead it will start up near-full screen, regardless of its previous setting.

Consider restoring the window size on startup as it was last set by the user.

Mingling94:
- Don't exit file adder in Files plugin unless a file uploads successfully so I (as the user) can try uploading again instead of redoing the steps
- Moved around files plugin code
- Upload directory detects .sia files and loads the file they point to instead of uploading the actual .sia
- Save size & position on ui close fixes #101


lukechampine:
needs rebase

Mingling94:
@lukechampine rebased

app/plugins/Files/js/loaders.js
```diff
@@ -76,7 +89,9 @@ function uploadDir(dirPath, nickname) {
 					notify('Cannot read ' + file, 'error');
 					return;
 				}
-				if (!isUnixHiddenPath(filePath) && stats.isFile()) {
+				if (file.slice(-4) === '.sia') {
+					loadDotSia(filePath);
```
lukechampine: I'm not sure this is a good idea. From the user's perspective, this behavior may be surprising. And what if they actually wanted to upload the .sia file?

Mingling94: What if one wants to upload a directory of .sia files though? Should it be a separate option like with uploading a single file?

lukechampine: I would think so. Or just modify `Add .sia File` to allow selecting multiple files.

Mingling94: Iunno... I think apps generally do expected conveniences like this. In what case would someone want to load a .sia file?

lukechampine: when in doubt, go with the least surprising behavior. Doesn't mean we can't have this functionality, just that it needs to be clear to the user.

PR #114, opened 2015-10-30
Usability fixes

Mingling94:
- Fixed error where the UI does not start if the config.json one currently has does not have a width, height, xPosition, and yPosition field. This checks for that and also updates the default config.json
- Fixed a long standing sidebar bug where one could select multiple views
- Gives option to save wallet password

### DIFF
--- a/app/js/pluginManager.js
+++ b/app/js/pluginManager.js
@@ -131,10 +131,12 @@ function PluginManager() {
 					break;
 				case 'config':
 					// get or set something in the config.json
-					// TODO: Document in PluginTutorial.md
 					var args = event.args[0];
 					var responseChannel = event.args[1];
-					plugin.sendToView(responseChannel, UI.config(args))
+					var result = UI.config(args);
+					if (responseChannel) {
+						plugin.sendToView(responseChannel, result);
+					}
 					break;
 				case 'devtools':
 					// Plugin called for its own devtools, toggle it
--- a/app/js/pluginManager.js
+++ b/app/js/pluginManager.js
@@ -26,11 +26,6 @@ function PluginManager() {
 	 */
 	var current;
 	/**
-	 * Stored password if option is set
-	 * @member {password} PluginManager~password
-	 */
-	var password;
-	/**
 	 * Array to store all plugins
 	 * @member {Plugin[]} PluginManager~plugins
 	 */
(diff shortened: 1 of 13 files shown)


Mingling94:
@DavidVorick @lukechampine boop

app/js/pluginManager.js
```diff
@@ -135,6 +129,16 @@ function PluginManager() {
 					event.args[1].left += $('#sidebar').width();
 					UI.tooltip.apply(null, event.args);
 					break;
+				case 'config':
+					// get or set something in the config.json
+					// TODO: Document in PluginTutorial.md
```
DavidVorick: should do this now

app/js/uiConfig.js
```diff
@@ -8,6 +8,10 @@
 const defaultConfig = {
 	homePlugin: 'Overview',
 	siadAddress: 'http://localhost:9980',
+	width: 800,
+	height: 600,
```
lukechampine, line 12: does this mean the UI will default to this size for all users? I'm so used to it being fullscreen haha

app/js/pluginManager.js
```diff
@@ -135,6 +129,16 @@ function PluginManager() {
...
+					var args = event.args[0];
+					var responseChannel = event.args[1];
+					var result = UI.config(args);
+					if (responseChannel) {
+						plugin.sendToView(responseChannel, result)
```
lukechampine: be diligent about ending lines with `;`.

app/js/uiConfig.js
DavidVorick, line 12: we definitely got a fair amount of complaints about the default size of the UI, people wanted it smaller

app/js/pluginManager.js
DavidVorick: [fails silently]

app/js/uiConfig.js
Mingling94, line 12: ^ It'll save upon adjustment though so w.e

app/js/pluginManager.js
Mingling94: fails silently?

app/js/pluginManager.js
Mingling94: I actually realized I had change the portion of the pluginManager where I detailed all the IPC channels and instead said to look at the code :P

app/js/pluginManager.js
lukechampine: if you were to insert another line here, the resulting behavior would be unpredictable. It could do what you intended, or it could break spectacularly, or it could fail silently. I believe JS now automatically inserts semicolons in certain scenarios, but I'd rather not risk it.

DavidVorick:
LGTM

though, the lack of a test suite makes me nervous

PR #117, opened 2015-11-04
Made css stylings more consistent

Mingling94:
I thought it odd that rows in the plugins were larger than the 60px height that were uniform to our buttons and headers so I made them all 60px.

### DIFF
--- a/css/general.css
+++ b/css/general.css
@@ -112,7 +112,7 @@ html,body {
 	background-color: #00CBA0;
 	color: #F5F5F5;
 }
-.button.current .icon{
+.button.current .icon, .button:hover .icon {
 	-webkit-filter: invert(100%);
 }
 @keyframes fadeout {
--- a/css/plugin-standard.css
+++ b/css/plugin-standard.css
@@ -121,3 +121,7 @@ select:hover {
 	position: relative;
 	width: 100%;
 }
+.button:hover {
+	color: #4a4a4a;
+	cursor: pointer;
+}
--- a/plugins/About/css/about.css
+++ b/plugins/About/css/about.css
@@ -10,11 +10,11 @@
  * 		Black:      #000000
  */
 .about {
-	color: #9e9e9e;
-	padding: 20px;
+	color: #c5c5c5;
+	padding: 15px 15px 14px 15px;
 	border-bottom: 1px solid #e0e0e0;
 }
 .about .version {
 	display: inline-block;
-	color: #757575;
+	color: #4a4a4a;
 }
--- a/plugins/About/css/about.css
+++ b/plugins/About/css/about.css
@@ -11,8 +11,8 @@
  */
 .about {
 	color: #c5c5c5;
-	margin-top: 5%;
-	margin-left: 5%;
+	margin: 5% 5% 0 5%;
+	padding-left: 10px;
 	padding-bottom: 10px;
 	border-bottom: 1px solid #e0e0e0;
 }
--- a/plugins/About/css/about.css
+++ b/plugins/About/css/about.css
@@ -11,7 +11,9 @@
  */
 .about {
 	color: #c5c5c5;
-	padding: 15px 15px 14px 15px;
+	margin-top: 5%;
+	margin-left: 5%;
+	padding-bottom: 10px;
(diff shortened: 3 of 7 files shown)


DavidVorick:
did this change the look at all? Can you post screenshots of before/after?

At some point, it would be nice to have some sort of framework that produced screenshots of the changes on various systems and resolutions. heh, more testing. (this is not a near-future need)

Mingling94:
It's minute, but rows are even in this branch and plugin buttons highlight to a uniform color:
sia_010
sia_022

As opposed to in master where the bars are misaligned compared to the header/sidebar buttons and the hover shading is defined per plugin and is slighlty lighter, as demonstrated by the `Add Folder` button.
sia_019
sia_015

DavidVorick:
I'm not sure that you want this stuff lined up. They are separate parts of the UI, and when they don't line up it's much more clear that there's a difference between them. Would be nice to get an opinion from a designer.

seveibar:
I could go wrong either way on this one, David makes a good point about different functional parts of the UI. But I like the appearance of the symmetry.

I think the symmetry might be better with a smallish border to distinguish the active area.

If on the fence, its probably a good idea to trust the in-house designer (Ming here) :)

DavidVorick:
@pmknutsen since you designed it, you're actually the in-house designer here

pmknutsen:
Some general comments (from a non-designer, but someone knowledgeable on visual processing):

Aligning elements has a visually calming effect. This is especially true when elements are perceived as belonging to the same 'group'. Our visual system groups by all kinds of features, including spatial features such as nearby edges with matching orientations. In this case, you have multiple horizontal lines that almost, but not quite, merge. So, we tend to perceive something continuous, because the elements almost line up. This triggers our attention to a perceived misalignment. In fact, the exact elements Ming aligned have also irked me on a few occasions. In a UI, you generally don't want a user's attention drawn to anything that is not a 'feature', i.e. you want a focus on application usability and not on design (i.e. a good design is something that does not draw any kind of attention away from the usability aspect).

Following this, a designer can also 'over-align'. This happens when you align elements that otherwise would not be visually grouped, e.g. because they are visually too distinct (e.g. don't share color, shape, orientation, too far apart etc).

I don't think it is important here that the menu and body of the UI are different 'parts' of the UI, and thus should not be aligned. You should align not because of function, but because you don't want to create a visual distraction if you otherwise did not align.

Since at least two people (Ming and I) were both irked by this small mismatch, I think that tells us that these elements should be aligned.

So, long story short, I too vote for alignment.

Mingling94:
Wow thanks for the insight @pmknutsen!

DavidVorick:
From Jing:

```
okay so
two design decisions i disagree with:
regardless of whether they line up or not, the bottom lines of the objects on the left should not extend all the way aross the page
across*
a little bit of lining up is ok
let me screenshot you an example

http://imgur.com/Q5pJwWC

search bar is aligned, but that's ok
site objects are not aligned
does that make sense?
```

image:
12214206_10153369626023795_568879534_o

DavidVorick:
Sorry, had a few meetings, didn't get a chance to read @pmknutsen's comment until now.

Jing's suggestions solve this problem by preventing the lines from going to the edge, which means your eyes don't try to line things up anymore

Mingling94:
So I think this is what Jing meant, specifically with the about page.
sia_026

DavidVorick:
I'm just nitpicking at this point, but since it's a style PR I guess that makes sense. I don't think you want the underline to line up perfectly with the text, if you look at Jing's screenshot, the line is actually larger than the text by a little bit.

Mingling94:
Yo, out of left-field here, but what about
sia_028

Otherwise this pr is updated with those slightly longer lines you wanted.
sia_029

DavidVorick:
Gotta say, now that I'm seeing it, I think it looks a lot better without the lines extended. I guess I'm not a designer lol.

I do like the idea of adding a text blurb with some links, but the way you did it above doesn't look good at all. Also, I'm not sure I like the version numbers in a capsule, seems like the wrong place to put them.

Mingling94:
Yeah that first screenshot is just for concept. Otherwise this PR looks like the second screenshot in my previous comment, can it be merged?

PR #122, opened 2015-11-05
Hosting

Mingling94:
Added button to specify address announcing

To be done in a future PR: Use Jquery across plugins where convenient and load up popup from a central html file to reduce redundancy


DavidVorick:
LGTM

PR #126, opened 2015-11-09
Lint

Mingling94:
Didn't have to insert jshint comments into code, just reorganized the code a bit.


js/.jshintrc
```diff
@@ -0,0 +1,98 @@
...
+		"Shell": false,
+		"Process": false,
+		"Remote": false,
+		"mainWindow": true,
+		"UI": true,
+		"Daemon": true,
+		"Plugins": true
+	},
+	"indent"        : 4         // Specify indentation spacing
+}
```
DavidVorick, line 98: are there two jshint files now?

DavidVorick:
had to reopen the PR to get travis to run, but looks like it's working now

js/.jshintrc
Mingling94, line 98: This is so index.js can be linted with different globals than the files in the js folder.

Also Files, Hosting, and Wallet all have their own .jshintrc's as well.

PR #132, opened 2015-11-17
Jqueried up the Wallet plugin

Mingling94:


js/daemonManager.js
```diff
@@ -105,7 +105,7 @@ function DaemonManager() {
 
 			// Listen for siad erroring
 			daemonProcess.on('error', function (error) {
-				if (error === 'ENOENT') {
+				if (error === 'Error: spawn ' + siaCommand + ' ENOENT') {
```
DavidVorick, line 108: yeah, fully agree with making the error messages more informative.

PR #133, opened 2015-11-18
Fix hosting plugin and wallet init endpoint

Mingling94:
Didn't realize I left hosting configuration to not show up.
Also wallet's `encrypt` endpoint is `init` with 0.4.8


Mingling94:
bloop @lukechampine @DavidVorick

PR #134, opened 2015-11-19
Wallet

Mingling94:
Wrote preliminary wallet testing.

Open to suggestions of what to test, leave it to me to figure out how.


plugins/Wallet/js/addresses.js
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// Make and append a single address
+function addAddress(address) {
+	makeAddress(address, appendAddress);
+}
+
+// From an array of addresses, make elements then insert into the page in a
+// semi-non-blocking manner
+// TODO: Could this be done better?
```
DavidVorick, line 61: I don't understand what is happening in this function

plugins/Wallet/js/addresses.js
```diff
@@ -0,0 +1,119 @@
...
+
+	/* Fetch all wallet transactions by iterate over wallet addresses
+	var loopmax = result.addresses.length;
+	var counter = 0;
+	(function next() {
+		setTimeout(function() {
+			updateAddrTxn(result.addresses[counter].address);
+			next();
+		}, 50); // force 50 ms delay between each GET request
+	})();*/
```
DavidVorick, line 87: why is this code commented out?

DavidVorick:
"Open to suggestions of what to test, leave it to me to figure out how."

In general, you want to
1. check that everything works as you expect it to work. If it's automatic, that saves a lot of human time lost when inspecting the product.
2. write tests so that it'll catch regressive changes in the future. This goes back into 1, but sometimes you already know that code works (from other testing). But it might not always work, because a dependency might change or another developer might start monkeying around and not realize that they've made a mistake. A failed test will let them know that they've done something that has introduced bugs, which is extremely helpful for keeping a quality codebase.

Mostly, I want to be able to look at a PR, see that the tests have all passed, and then have a reasonable confidence that they haven't broken anything too severely. For UI, this is pretty hard especially because we have to be careful about different resolutions, etc.

In general though, everything that needs to work should have a test. All code should have a purpose, and there should be a test that checks the purpose is being satisfied according to specification.

plugins/Wallet/js/addresses.js
Mingling94, line 61: It passes the address to makeAddress() to make the html element, sends the result to appendAddress() to add that html element to the list.

This will likely be redone to paginate viewing of addresses and transactions.

plugins/Wallet/js/addresses.js
Mingling94, line 87: https://github.com/NebulousLabs/Sia-UI/pull/80#discussion_r41283414

PR #135, opened 2015-11-24
Paginated Addresses and Address Search

Mingling94:
Addresses now display in pages of 25 results each. Search functions the same.

### DIFF
--- a/test/renderer.js
+++ b/test/renderer.js
@@ -51,7 +51,7 @@ describe('renderer process', function() {
 			// script to execute in the UI
 			function script(addresses) {
 				function sendAddresses() {
-					Plugins.Wallet.sendToView('update-address', null, {
+					Plugins.Wallet.sendToView('update-addresses', null, {
 						addresses: addresses,
 					});
 				}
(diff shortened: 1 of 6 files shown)


DavidVorick:
you don't want to get rid of tests, you want to fix them so that they either pass every time or so that whatever bug was causing them to fail is fixed. More often than not, the reason they fail on occasion is because there's a bug

test/renderer.js
```diff
@@ -83,28 +83,8 @@ describe('renderer process', function() {
 		it('wait until loaded', function() {
 			return client.waitUntilWindowLoaded();
 		});
-		it('appends 1 addresses', function(done) {
-			addNAddresses(1, done);
-		});
-		it('appends 100 addresses', function(done) {
-			addNAddresses(100, done);
-		});
```
DavidVorick: this fails nondeterministically?

Mingling94:
Honestly the tests were just so I could manually see what lots of addresses loaded in the plugin looked like. I'm writing more intelligent tests now that are actually automatic and aren't just useful for manual testing.

PR #141, opened 2015-12-01
Misc

Mingling94:
- Plugin devtools hotkey now does its job
- Some css changes, mostly to reduce redundancy
- Update checking for the UI via Github API
- Most notification icons (except error and success) are neon green!

### DIFF
--- a/js/uiManager.js
+++ b/js/uiManager.js
@@ -252,9 +252,6 @@ function UIManager() {
 			Daemon.init(config);
 			Plugins.init(config);
 		});
-		$('#update-button').click(function() {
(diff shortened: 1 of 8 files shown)


js/uiManager.js
```diff
@@ -237,11 +238,38 @@ function UIManager() {
 		}, 2500);
 	};
 
+	// Checks if there is an update available
+	function checkUpdate() {
+		$.ajax({
+			url: 'https://api.github.com/repos/NebulousLabs/Sia-UI/releases',
```
lukechampine, line 244: awesome :+1:

PR #143, opened 2015-12-03
A practice in object-oriented Node.js

Mingling94:
- Removed daemonAPI.js
- Require js managers in the index.html rather than DOM src tags
- Capitalized public members of closures


lukechampine:
merge conflicts

Mingling94:
In order to make Files a viable files browser that isn't a clusterfuck of global namespace js, I'll have to modularize it into an OOP setup. The need to send and listen on an IPC call for every API call makes this quite messy.

I'm adding a bunch of stuff and refactoring in order to make daemonManager completely separate, and thus `require`-able from the plugins.

Thus, revisit this later when I'm more done with it, look at #144 first

PR #142, opened 2015-12-02
Siafund

Mingling94:
On top of #140 and some of #141 

Can view transactions, siacoin or siafund. Can also make siafund transactions now.

TODO after this PR:
- Add wallet settings page with the following:
  - Filter what transactions one views
  - 'Siafund' mode switch that unlocks this so one can't accidentally send siafunds. Perhaps even have another mode switch in the general UI so one REALLY has to be sure of sending siafunds.
  - Load Legacy Wallet button

PR #144, opened 2015-12-04
Wallet fixup

Mingling94:
On top of #142 
Here's what the plugin looks like half screen and whole screen:
sia_044
sia_043

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "doc": "jsdoc -c .jsdocrc",
     "sia-repo": "go get -u github.com/NebulousLabs/Sia/...",
     "sia-import": "(test -d Sia || mkdir -p Sia) && (cp $GOPATH/bin/siad Sia & cp $GOPATH/bin/siac Sia)",
-    "sia-kill": "killall siad; while pgrep siad; do sleep .5; done;",
+    "sia-kill": "killall -s SIGKILL siad; while pgrep siad; do sleep .5; done;",
     "sia-refresh": "npm run sia-kill && rm -rf Sia && npm run sia-import",
     "release-args": "electron-packager . Sia-UI --app-version=0.4.8-beta --version=0.34.1 --overwrite --prune --ignore=release --out=release/v0.4.8-beta --icon=app/assets/icon",
     "release": "npm run sia-import ; npm run release-args -- --all && ./.finalizeRelease.sh",
--- a/package.json
+++ b/package.json
@@ -10,11 +10,11 @@
     "chai-as-promised": "^5.1.0",
     "electron-mocha": "^0.6.1",
     "electron-packager": "^5.1.1",
-    "electron-prebuilt": "^0.35.0",
+    "electron-prebuilt": "^0.35.4",
     "ink-docstrap": "^0.5.2",
     "jsdoc": "^3.3.3",
     "jshint": "^2.8.0",
-    "spectron": "^0.34.0"
+    "spectron": "^0.35.3"
   },
   "dependencies": {
     "bignumber.js": "^2.1.0",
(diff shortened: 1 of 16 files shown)


lukechampine:
The red seems kinda bright. Is there a corresponding green for when you received money?

Mingling94:
There is, they are all-red or all-green, meaning hex-values #FF0000 or #00FF00, I can tone them down a bit to like #aa0000 and #00aa00 if you'd like.

pmknutsen:
Why remove the copy-to-clipboard shortcut in Addresses? I find it very convenient.

pmknutsen:
And, was the Create Address removed as well, or moved elsewhere?

Mingling94:
@pmknutsen I enabled a right-click context-menu a while ago with the whole UI that allows copy-paste, but I could add that back in. Create Address is the "+" button in the lower right.

DavidVorick:
I am in favor of bringing back the copy-to-clipboard button

Mingling94:
Copy-to-clipboard button is back!
sia_045

Mingling94:
@lukechampine @DavidVorick boop

package.json
```diff
@@ -29,7 +29,7 @@
     "doc": "jsdoc -c .jsdocrc",
     "sia-repo": "go get -u github.com/NebulousLabs/Sia/...",
     "sia-import": "(test -d Sia || mkdir -p Sia) && (cp $GOPATH/bin/siad Sia & cp $GOPATH/bin/siac Sia)",
-    "sia-kill": "killall siad; while pgrep siad; do sleep .5; done;",
+    "sia-kill": "killall -s SIGKILL siad; while pgrep siad; do sleep .5; done;",
```
lukechampine: Don't send `SIGKILL`. `SIGTERM` gives siad a chance to shutdown cleanly. Sending `SIGKILL` could cause corruption.

plugins/Wallet/css/lists.css
```diff
@@ -0,0 +1,99 @@
...
+/* Transactions list */
+.transaction > * {
+	overflow: hidden;
+}
+.currency.positive {
+	color: #00FF00;
+}
+.currency.negative {
+	color: #FF0000;
+}
```
lukechampine, line 48: yeah, these are too bold. Tone them down a bit.

plugins/Wallet/js/transactions.js
```diff
@@ -92,13 +92,13 @@ $('#transaction-page').on('input', updateTransactionPage);
...
-	       (criteria.maxValue  === undefined || txn.value                 <= criteria.maxValue)  &&
-	       (criteria.minValue  === undefined || txn.value                 >= criteria.minValue)  &&
-	       (criteria.currency  === undefined || criteria.currency.indexOf(txn.currency) !== -1)  &&
-	       (criteria.inputs    === true      || txn.value                 >= 0)                  &&
-	       (criteria.outputs   === true      || txn.value                 <= 0);
+	return (criteria.endTime   === null || txn.confirmationtimestamp <= criteria.endTime)   &&
+	       (criteria.startTime === null || txn.confirmationtimestamp >= criteria.startTime) &&
+	       (criteria.maxValue  === null || txn.value                 <= criteria.maxValue)  &&
+	       (criteria.minValue  === null || txn.value                 >= criteria.minValue)  &&
+	       (criteria.currency  === null || criteria.currency.indexOf(txn.currency) !== -1)  &&
```
lukechampine, line 99: what's the advantage of switching to `null` here? I assume it protects against using non-existent fields, e.g. `criteria.foo`?

Mingling94, line 99: No advantage, there are only slight differences to them. It just felt wrong to initialize something, or `define` it, as `undefined`.

Rather a null criteria just means it's relaxed and won't limit the transaction search.

lukechampine:
for future reference, better than amending here would be to specifically `git revert c219021` to make your intentions clear.

DavidVorick:
The eyeglass is too close to the addresses, it looks unnatural

Mingling94:

PR #146, opened 2015-12-08
Hosting

Mingling94:
On top of #145 

Updated the Hosting plugin to the new API. While I was at it, I wanted to make the plugin as OOP as possible. This was enabled by requiring the daemonManager and no longer relying on IPC for api calls.

This demonstrates that the daemonManager could be successfully used as an npm package.


lukechampine:
rebase pls

Mingling94:
@lukechampine oki didit

plugins/Hosting/js/host.js
```diff
@@ -0,0 +1,118 @@
+'use strict';
+
+// Manges the lifecycle of the plugin
+function host() {
+	// Tracks details about the various hosting properties
+	var props = require('./hostData.js');
+	// Hold Sia math logic 
+	var S = require('./siaMath.js');
```
lukechampine: `S` doesn't imply anything about math...

plugins/Hosting/js/siaMath.js
```diff
@@ -0,0 +1,68 @@
...
+// Display # of decimals
+const DISPLAY_DECIMALS = 2;
+
+// Catch if BigNumber, property object, or anything else is passed in
+function interpret(thing) {
+	if (typeof thing === 'object') {
+		thing = thing.value;
+	}
+	return new BigNumber(thing).toNumber();
+}
```
lukechampine: this file seems a little "fast and loose," if you get what I mean. Functions like `interpret`, `convert`, and `revert` are generic to the point that I can't easily get a sense of what they do and why. Basically, there's a lot of type information missing here. `formatSiacoin` and `formatBytes` are good because they specify the type. What are the rest for? Generic conversions to/from bignums?

Also, just a personal preference, but I think it's cleaner to define the functions at top-level, and then have a compact `module.exports` at the bottom.

Mingling94: It's important to read them with the context of hostData.js, But I'll try to make the file more explicit

PR #147, opened 2015-12-09
Sia.js

Mingling94:
I published 0.1.0 of sia.js, the node wrapper for siad, earlier. This pr uses it in place of daemonManager.js


Mingling94:
Going to have to wait for #146 to convert the Hosting plugin

plugins/Overview/js/overview.js
```diff
@@ -40,6 +40,42 @@ function formatSiacoin(hastings) {
...
+	var unlocked = result.unlocked;
+	var unencrypted = !result.encrypted;
+
+	var lockText = unencrypted ? 'New Wallet' : unlocked ? 'Unlocked' : 'Locked';
+	document.getElementById('lock').innerText = lockText;
+
+	var bal = unlocked ? formatSiacoin(result.confirmedsiacoinbalance) : '--';
+	document.getElementById('balance').innerText = 'Balance: ' + bal;
+}
+function updatePeers(err, result) {
```
DavidVorick, line 60: each function should have a newline and a docstring

Mingling94:

lukechampine:
looks like a nice cleanup, my only concern is with the multiple calls to `Siad.configure`. Can you explain each of them?

Mingling94:
Each plugin imports/`require`s an instance of the siad wrapper. They ask the UI through ipc for the siad settings:

``` json
"siad": {
        "path": "/home/ming/Github/ui/Sia",
        "address": "http://localhost:9980",
        "command": "siad",
        "headers": {
            "User-Agent": "Sia-Agent"
        },
        "running": false
    }
```

To update their own wrapper instances with. None of these properties are really needed unless the user hosts siad on a different address or some other custom setting that would then propagate to the plugins.

lukechampine:
ah. Seems a little awkward to have to replicate that everywhere `Siad` is used...maybe we can think of a better solution.

PR #154, opened 2015-12-14
Golang no longer a prequisite, got rid of globally required imports

Mingling94:
Various changes that _seem_ unrelated but actually are. Explanation in chronological order:
- Fixed `.finalizeRelease.sh` as it wasn't putting the correct Sia release into each Sia-UI release before
  - This took a bunch of inspiration from David's previous release.sh code
- From there I saw that Golang wasn't needed for anything else besides `npm run sia-import` and so I added `justGetSia.js` to download sia on post-install instead of relying on the user to have a `$GOPATH/bin/siad`
- Fixed bug with how `siad.configure` was being called. Irrelevant with #155 but twas a quick fix.
- Because jshint was complaining about the `justGetSia.js` redefinition of the globally require 'path' node library, I got rid of global imports on the general UI level. They seemed like more trouble than they were worth anyway.

### DIFF
--- a/.finalizeRelease.sh
+++ b/.finalizeRelease.sh
@@ -1,35 +1,84 @@
 #!/bin/bash
 
 set -o errexit # Exit on error
-readonly UI_VERSION="v0.4.8-beta"
+readonly VERSION="v0.4.8-beta"
+declare -a Electron_Archs=("linux-x64" "linux-ia32" "win32-x64" "win32-ia32" "darwin-x64")
+declare -a Sia_Archs=("linux-amd64" "linux-386" "windows-amd64" "windows-386" "darwin-amd64")
 
-cd "release/$UI_VERSION"
+cd "release/$VERSION"
 
-# Array of different distribution strings
-declare -a OSes=("linux-x64" "linux-ia32" "win32-x64" "win32-ia32" "darwin-x64")
-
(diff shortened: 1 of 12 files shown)


.finalizeRelease.sh
```diff
@@ -1,35 +1,84 @@
...
-# Array of different distribution strings
-declare -a OSes=("linux-x64" "linux-ia32" "win32-x64" "win32-ia32" "darwin-x64")
+# generic packaging function
+package() {
+	local electron_arch="$1"
+	local ui="Sia-UI-$electron_arch"
+	local extension=".zip"
+	if [ "${1:0:5}" == "linux" ]; then
+		extension=".tar.gz"
+	fi
```
lukechampine, line 17: declaring the extension here seems unnecessary since you branch on arch anyway

Mingling94, line 17: It makes it so I can do the following series:

```
local sia_archive="${sia_folder}${extension}"
local sia_url="https://github.com/NebulousLabs/Sia/releases/download/${VERSION}/${sia_archive}"
```

to get the one-line of wget (instead of 3):

``` bash
wget -q $sia_url
```

Mingling94:
@DavidVorick @lukechampine @VoidingWarranties boop

PR #155, opened 2015-12-15
Reorganization of the main process

Mingling94:
On top of #154 
Major things:
- Main process code separated out into parts and put into new folder `Sia-UI/mainjs` to distinguish from `Sia-UI/js`. **I'm open to a better naming scheme.**
- Moved Config management logic to main process. It seems more logical and makes getting and setting the config.json through ipc much simpler.
  Turns

```
(plugin sends ipc to host/UI) ->
(pluginManager receives ipc and calls uiManager config) ->
(uiManager gets or sets and returns) ->
(pluginManager receives and sends to plugin) ->
(plugin listens on ipc)
```

into

```
(plugin sends synchronous ipc to main process) ->
(main process gets or sets and returns) ->
(plugin receives value without listening on an ipc channel)
```
- Siad startup logic now in main process as well.
- As a result, siad closes upon UI exiting. It didn't before on Linux even though it was supposed to.
- Upon startup, if Sia is not detected, a series of dialog messages are shown for the user to either 1. Pick the location of their siad file or 2. Download one into their root folder

Minor things:
- `npm install` runs `npm run sia-import`. Not sure why I took that out before.
- Moved notification logic from uiManager to another js file
- Revamped the notification system a bit
- Fixed a bunch of notification errors before. It was pretty broken before
- Changed `notify` ipc channel to `notification` everywhere


DavidVorick:
justGetSia.js is not a great name, what about getSia.js?

DavidVorick:
what is the difference between the `mainjs` and `js` folders?

PR #161, opened 2015-12-16
Removed global variables in Hosting plugin

Mingling94:
I didn't like how each file needed to know what variables were global and how I
needed to maintain a .jshintrc. Instead I have no global variables and
import/`require` everything needed per file.

It's okay to `require` the same file from two different ones because node.js
expects this. When a file `require`s a module or file that's been `require`d
elsewhere, it gets the same instance as before. The `require`d object is cached.

This doesn't apply to `require`-ing something from two different processes, as in
from a webview and from the main process, so plugins can't share the same
config instance with each other or the main process.

This also made circular dependencies more obvious so I had to add some
callbacks to deal with resulting errors


DavidVorick:
Though I think this is a good change overall we really need to finish the changes to the renter. I would much rather see a PR related to that than incremental improvements to plugins such as the host, which in my mind are already 'good enough'. We need to keep focused on the release schedule.

Mingling94:
Gotcha. Most of my working time is spent researching oop js so as to write code for the file browser that doesn't need as much refactoring later. This actual code took 1-2 hours.

The important conceptual part of this PR is the note about modules being cached when `require`d.

PR #162, opened 2015-12-31
Facilitate a config.siad.detached option

Mingling94:
Not only will some users/devs want siad as a detached process, but it
allows easier and faster debugging since one won't have to wait siad to
load every time

Corresponds to https://github.com/NebulousLabs/Nodejs-Sia/pull/14
**NOTE**: the Nodejs-Sia PR should be accepted first, version bumped, and then published before this PR can actually utilize those changes.

### DIFF
--- a/js/mainjs/initSiad.js
+++ b/js/mainjs/initSiad.js
@@ -131,9 +131,8 @@ function checkSiadPath() {
 		if (!siadPath) {
 			checkSiadPath();
 		} else if (selected === 0) {
-			var lastIndex = siadPath.lastIndexOf('/');
-			config.siad.command = siadPath.substring(lastIndex);
-			config.siad.path = siadPath.substring(0, lastIndex);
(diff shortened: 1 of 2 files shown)


js/mainjs/initSiad.js
```diff
@@ -132,7 +132,7 @@ function checkSiadPath() {
 			checkSiadPath();
 		} else if (selected === 0) {
 			var lastIndex = siadPath.lastIndexOf('/');
-			config.siad.command = siadPath.substring(lastIndex);
+			config.siad.fileName = siadPath.substring(lastIndex);
```
DavidVorick: what's going on here?

Mingling94: Oh I renamed `siad.command` to `siad.fileName` in `sia.js` of the node wrapper and forgot to reflect that change here.

PR #163, opened 2016-01-02
Revamped Files plugin

### LINKED ISSUE #158: File plugin updating
The file view should update immediately after a file is added or deleted. (The latter is especially confusing.)

Mingling94:
On top of #162 sort of. That PR is awaiting changes in a sia wrapper PR: https://github.com/NebulousLabs/Nodejs-Sia/pull/14

**Sorry this PR has taken a while, but most individual commits until this point in the branch left the plugin broken or really error-prone, as would be the case with a large redesign.**
## Major changes:
- Files plugin has classes to be semi-oop!
  - Of special note are the classes `entity`, `folder`, `file` which model a virtual file system for a renter's library. `entity` isn't used except as the parent class of `folder` and `file` (and perhaps more?)
- Jqueried up Files plugin! Allows for some nifty transitions!
  - Got rid of html blueprints (hidden divs for cloning purposes) and now create elements from html-strings such as within `fileElement.js` and `folderElement.js`
- Changed how files are added to a dynamic dropdown menu, inspired by https://drive.google.com/drive/my-drive
  - Files and folders are uploaded as is, reducing the need for popups/divs asking for optional nicknames. TODO: implement easy renaming and drag&drop file/folder movement
## Minor changes:
- For this plugin only, I've messed around with styling and naming practices. They all still pass JSHint linting, but are different from how I do the rest of the UI. However, after my experiences reading other JS repos, I believe they are safe and more conventional to adopt. These changes include:
  - Only capitalize class constructors (which I don't really use since I make classes via [`Object.assign(Object.create(superclass), subclass);`](https://medium.com/javascript-scene/common-misconceptions-about-inheritance-in-javascript-d5d9bab29b0a)
  - Various es5 syntax sugars
- Actively looked for places to remove css. Things may look a bit ugly at first but css maintenance seemed slightly useful at best and arbitrary time waste at worst unless it is guaranteed to go into the next release.
## TODO's:
- File/folder drag&drop movement (as mentioned earlier)
  - Multi-select actions?
- Easy renaming
- Get search to search through all files/folders under the current folder
- Hotkeys to commonly used actions
## Screenshots:

sia_047
sia_048
selection_050
selection_051
### The above shows this PR fixes #158

sia_052
sia_053


Mingling94:
@DavidVorick @lukechampine @VoidingWarranties 
I invested quite a bit of time into this and I'd appreciate a thorough code review. Especially consider that I'm experimenting with some new JS conventions/tools/architecture things that I'm totally open to input/criticism on.

I understand there's 2k+ line changes and that might take a while. I'll be developing on top of this PR in another branch.

DavidVorick:
Great to see this, thanks. I will try to look at it asap, though I can't promise too much in terms of quality because I'm not very good with javascript. I will give it my best though

lukechampine:
Maybe it's just unclear from the screenshots, but can you nest folders within folders? If so, we'll need some way of displaying the user's current path within the directory structure. Google Drive again seems like a good place to draw inspiration from.

Mingling94:
@lukechampine The part where it says `New Folder` is meant to be a CWD just like with GDrive and could be `New Folder/New Folder`. I just haven't gotten it working yet. That's next up and reserved for another PR on top of this one.

.jshintrc
```diff
@@ -33,7 +33,6 @@
 	"boss"          : false,    // Tolerate assignments inside if, for & while. Usually conditions & loops are for comparison, not assignments.
 	"debug"         : false,    // Allow debugger statements e.g. browser breakpoints.
 	"eqnull"        : false,    // Tolerate use of `== null`.
-	"es5"           : false,    // Allow EcmaScript 5 syntax.
```
DavidVorick, line 36: in the future, better to have a change like this be a separate PR.

js/mainjs/addIPCListeners.js
```diff
@@ -30,7 +34,10 @@ module.exports = function(config, mainWindow) {
 			default:
 				console.error('Unknown dialog ipc');
 		}
-		event.returnValue = response ? response : null;
+		// TODO: Make async and adapt plugins
```
DavidVorick, line 38: can you explain this todo a bit better? I'm guessing there's not going to be enough time to do this.

.jshintrc
Mingling94, line 36: ES5 standards are used heavily throughout the PR

js/mainjs/addIPCListeners.js
Mingling94, line 38: It's not that hard. Currently dialog windows block the UI, a minor annoyance but makes for slightly easier to read code. From the [dialog module in the electron documentation](http://electron.atom.io/docs/v0.36.0/api/dialog/):

> If a callback is passed, the API call will be asynchronous and the result will be passed via callback(filenames)

pmknutsen:
Nice job on the plugins! I look forward to testing this in a few days. I have some comments on the screenshots:

"New" context-menu:
Display menu below and left-aligned with New. it should not cover New. I think the plus is redundant.

For consistency, I suggest to style popup menus as popup notifications: straight, gray borders w/shadows. Horizontal dividers in popups should be a fainter gray than text.

Fix popup notification clutter:
When selecting multiple files/folders for upload, display a _single_ popup saying: "Added N files".

Feature request:
Deleting multiple files is a hassle (click one, confirm one, repeat, repeat, repeat...). Options:
1) Click on a filename, highlight it and remove with Delete key.
2) Instead of delete icons, display a checkbox after each file and a Delete option in the plugin toolbar.

Files / folders list:
1) Display folders before files.
2) Sort files alphabetically (later, add option to sort by filesize, time remaining like a conventional file browser).  

When user is not in the root folder, display a back arrow rather than folder icon in toolbar.

Throughout, I think text in the plugin frames need to be a bit darker. Text appears faint on some monitors.

Mingling94:
Aight, the PR is at a fairly stable point now, though I'm not trying to put forth many more hours to it today. I referenced many core things still yet TODO: in this issue: https://github.com/NebulousLabs/Sia-UI/issues/174

PR #164, opened 2016-01-03
Use sia.js 0.1.4 with detached option

Mingling94:
This should have been a part of #162 but it was merged right after https://github.com/NebulousLabs/Nodejs-Sia/pull/14 was.


Mingling94:
I'm just going to merge this immediately...

PR #165, opened 2016-01-05
Adapted to api changes in recent Sia pr

Mingling94:
See: https://github.com/NebulousLabs/Sia/pull/891/files

### DIFF
--- a/plugins/Hosting/js/host.js
+++ b/plugins/Hosting/js/host.js
@@ -46,11 +46,13 @@ function buildProps(hostInfo) {
 function update(hostInfo) {
 	// Store hostInfo results property data
 	$.each(hostInfo, function(name, value) {
+		// Try-catch statement to be fault-tolerant. If just one name from the
+		// api call result, hostInfo, isn't in hostProperties or has a
+		// different name, such as with an unexpected siad version, other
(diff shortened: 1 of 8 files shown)


plugins/Hosting/js/host.js
```diff
@@ -46,7 +46,15 @@ function buildProps(hostInfo) {
...
-		props[name].value = value;
+		try {
+			props[name].value = value;
+		} catch (e) {
+			// 'name' from hostInfo probably didn't match a member of the props
+			// object
+			console.error(e);
+			console.log('Error updating host properties object:');
+			console.log(hostInfo, props);
+		}
```
DavidVorick, line 59: Why does there need to be a try-catch here? The comment isn't very helpful.

lukechampine, line 59: Maybe the dereference can fail if `props[name]` doesn't exist?

Mingling94, line 59: It makes it a bit more fault-tolerant. I noticed that if just one `name` from the returned JSON, `hostInfo`, was off, all properties wouldn't update. This makes it so only that one property doesn't update and the host still works for the most part.

PR #166, opened 2016-01-05
Changed how plugin's start/stop

Mingling94:
In hindsight, executing start() and stop() for the plugin is more
confusing than sending ipc that it is being hidden/shown

### DIFF
--- a/plugins/About/js/about.js
+++ b/plugins/About/js/about.js
@@ -4,11 +4,16 @@
 const IPCRenderer = require('electron').ipcRenderer;
 // Siad wrapper
 const Siad = require('sia.js');
+
 // Make sure Siad settings are in sync with the rest of the UI's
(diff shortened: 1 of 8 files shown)


plugins/About/js/about.js
```diff
@@ -4,11 +4,16 @@
 const IPCRenderer = require('electron').ipcRenderer;
 // Siad wrapper
 const Siad = require('sia.js');
+
 // Make sure Siad settings are in sync with the rest of the UI's
 var settings = IPCRenderer.sendSync('config', 'siad');
 Siad.configure(settings);
 // Keeps track of if the view is shown
 var updating;
+const refreshRate = 50000;
```
lukechampine: why does the About page need to refresh at all?

Mingling94: True dat, it's mostly for conformity. I'll reduce it down.

PR #168, opened 2016-01-06
Fixed wallet locking problems

Mingling94:
Not exactly sure why, but a '/wallet' call to siad through the node.js wrapper immediately after  locking/unlocking the wallet wouldn't return. I could console.log evidence right before the call was being made within sia.js, but not after.

I found wrapping the call in a `setTimeout(funct, 1000)` seemed to allow this call to work, yet `process.nextTick(funct)` didn't. Then I reduced further to `setTimeout(funct, 0)` and then `setImmediate(funct)`.

An explanation of the difference between `process.nextTick` and `setImmediate` can be found [here](http://stackoverflow.com/a/15349865/1026909)

Asynchronous programming...

### DIFF
--- a/js/mainjs/config.js
+++ b/js/mainjs/config.js
@@ -33,7 +33,10 @@ function configManager(path) {
 	 */
 	config.attr = function(key, value) {
 		if (value !== undefined) {
-			this[key] = value;
+			config[key] = value;
+		}
+		if (config[key] === undefined) {
+			config[key] = null;
 		}
 		return config[key];
 	};
--- a/plugins/Wallet/js/locking.js
+++ b/plugins/Wallet/js/locking.js
@@ -6,26 +6,20 @@ var wallet = {};
 // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ Lock Icon  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 // Helper function for the lock-icon to make sure its classes are cleared
 function setLockIcon(lockStatus, iconClass) {
-	$('#lock-status').html(lockStatus);
-	$('#lock-icon').get(0).className = 'fa ' + iconClass;
+	$('#lock-pod span').text(lockStatus);
+	$('#lock-pod .fa').get(0).className = 'fa ' + iconClass;
 }
 
-// Markup changes to reflect locked state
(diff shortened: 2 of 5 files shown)


DavidVorick:
oy vey, glad you found the issue

js/mainjs/config.js
```diff
@@ -33,7 +33,10 @@ function configManager(path) {
 	 */
 	config.attr = function(key, value) {
 		if (value !== undefined) {
-			this[key] = value;
+			config[key] = value;
+		}
+		if (config[key] === undefined) {
+			config[key] = null;
```
DavidVorick, line 39: I don't understand this pattern. Is this fixing a problem that you found, or is it just defensive programming? (there's nothing wrong with defensive programming, I'm just trying to understand what the change is doing)

plugins/Wallet/js/locking.js
```diff
@@ -59,7 +53,9 @@ function updateStatus(result) {
 
 // Make wallet api call
 function getStatus() {
-	Siad.apiCall('/wallet', updateStatus);
+	// TODO: Don't understand why the setImmediate is needed, but without it,
+	// the '/wallet' call seems to not return right after a locking/unlocking
```
DavidVorick, line 57: huh. Isn't `updateStatus` supposed to only be called after the result of the api call has been returned? Are you double calling a function somewhere, or stacking events? Doesn't look like it, but I don't have a lot of ideas.

js/mainjs/config.js
Mingling94, line 39: This is for the line in `addIPCListeners.js` because a significant error (it throws a popup instead of just logging stuff) occurs:

``` js
event.returnValue = config.attr(key, value);
```

`event.returnValue` apparently can't be `undefined`

It indicates when a `key` is asked for and it isn't defined, it's set to `null` which isn't bad. A settings plugin would predictably blow up the config.json into a html form where values are editable, thus a `walletPassword` key with a null value could be edited and distinguished from an unknown key.

plugins/Wallet/js/locking.js
Mingling94, line 57: I don't know exactly what you mean by 'stacking events' but perhaps. It's not being double-called as far as I know.

PR #170, opened 2016-01-07
`qs` is less understandable than `form`

Mingling94:


DavidVorick:
So, my understanding of RESTful APIs, you want to be using a querystring for GET requests, and you want to be using body for POST requests. 'form' should only be used if you are using html forms, I'm not sure why this needs to be form, and why just using 'body' isn't enough.

I would feel more comfortable if we could figure out why `body` isn't sufficient. Maybe it's because you need to parse things into a string before passing it to the body, but for a form, the backend will automatically do the conversion?

Mingling94:
I'll look into why `body` isn't working

PR #171, opened 2016-01-08
Documentation

Mingling94:
This is in progress documentation, I just wanted to see if anyone has any input on how I'm writing it. It's probably best to just view the files and not worry about the line diffs.


DavidVorick:
Looks  alight to me. All of the linking to guides and such is definitely a positive

PR #172, opened 2016-01-08
Fixed dialog. It sometime blocked renderer process

Mingling94:
with no returnValue, a synchronous ipc call never returned

### DIFF
--- a/js/mainjs/addIPCListeners.js
+++ b/js/mainjs/addIPCListeners.js
@@ -34,7 +34,11 @@ module.exports = function(config, mainWindow) {
 			default:
 				console.error('Unknown dialog ipc');
 		}
-		response = response === undefined ? null : response;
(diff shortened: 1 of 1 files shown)


js/mainjs/addIPCListeners.js
```diff
@@ -30,7 +30,9 @@ module.exports = function(config, mainWindow) {
 			default:
 				console.error('Unknown dialog ipc');
 		}
-		event.returnValue = response ? response : null;
+		response = response === undefined ? null : response;
```
DavidVorick: I think that this statement is complex enough to merit an if statement.

``` js
if response === undefined {
     response = null;
}
```

DavidVorick:
Merge conflicts

Mingling94:

Issue #175, opened 2016-01-10
File plugin complaints


lukechampine:
- [x] Renaming should be done via a button, not double-clicking. Double clicking should always download a file (or enter a folder). Currently the hitboxes for file/folder names is far too large, extending all the way to the right.
- [x] The download/share/delete buttons do not work at all. They fail with the error `Cannot read property 'count' of undefined`.
- [x] Updating the DOM clears the user's selections.
- [x] Navigating folders is unintuitive. The action dropdown should always be triggered from the same place (perhaps the folder icon, or a `+` to the right of the path), rather than the current folder name.
- [x] When adding a file to a newly-created folder "foo", the file is actually uploaded to folder "undefined".
- [x] Renaming an undefined folder does not update the folder in siad.

PR #176, opened 2016-01-12
Files plugin fixes/adaptation

Mingling94:
- Adapted to new API
- Fixed various bugs (Still addressing many concerns of #175 though)
- Share and shareascii for folders
- Also fixed some deprecated api usage (file.nickname -> file.siapath)
  - Reflect new `renew` attribute in files and file elements
- tools.notType(), function to log improper arguments
  - Check param in each file/folder function
- Removed file.size and file.count getters, trying not to cater file class to folder class. Folder should strictly inherit from file
- Name fields do not take up more space than needed

Note: Many buttons still not working. Much `TODO:`

### DIFF
--- a/plugins/Files/js/browser.js
+++ b/plugins/Files/js/browser.js
@@ -113,12 +113,25 @@ function updateCWD(navigateTo) {
 	folders.push(currentFolder);
 
 	// Add a directory element per folder
-	folders.forEach(function(f) {
+	folders.forEach(function(f, i) {
 		var el = $(`
 			<span class='button directory' id='dir-${f.path}'>
-				${f.name}/
 			</span>
 		`);
+		// Root folder
+		if (f.path === '') {
+			el.html('<i class=\'fa fa-folder\'></i>');
+		} else {
+			// Middle folders
+			el.html(f.name);
+		}
+
+		// Last folder
+		if (i === folders.length - 1) {
+			el.append(' <i class=\'fa fa-caret-down\'></i>');
+		} else {
+			el.append('/');
+		}
 
 		// Clicking the element navigates to that folder
 		el.click(function() {
@@ -127,9 +140,6 @@ function updateCWD(navigateTo) {
 
 		// Append and add icon for root folder
 		cwd.append(el);
-		if (f.path === '') {
-			el.prepend('<i class=\'fa fa-folder\'></i>');
-		}
 	});
 
 	// Sum up widths to move dropdown right as directory is deeper
@@ -140,7 +150,7 @@ function updateCWD(navigateTo) {
 	});
 
 	// New file/folder button appears below last folder
-	cwd.children().last().click(function() {
+	cwd.children().last().off('click').click(function() {
 		var dropdown = $('.hidden.dropdown');
 		dropdown.css('left', cwdLength + 'px');
 		dropdown.toggle('fast');
--- a/plugins/Files/js/browser.js
+++ b/plugins/Files/js/browser.js
@@ -62,42 +62,38 @@ function getSelectedElements() {
 
 // Returns selected files/folders from currentFolder
 function getSelectedFiles() {
-	return getSelectedElements().map(el => currentFolder.files[el.id]);
+	return getSelectedElements().map(el => currentFolder.files[el.find('.name').text()]);
 }
 
 // Refresh the file list according to the currentFolder
 // TODO: folders before files, sort alphabetically
 function updateList(navigateTo) {
-	var list = $('#file-list');
+	// Get rid of all elements that don't belong
+	var files = currentFolder.filesArray;
+	var hashes = files.map(file => file.hashedPath);
+	$('.file:not(.label)').each(function() {
+		if (hashes.indexOf(this.id) === -1) {
+			$(this).remove();
+		}
+	});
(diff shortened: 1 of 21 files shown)


plugins/Files/js/file.js
```diff
@@ -98,42 +108,47 @@ var file = {
...
+	// Misc. functions
+	get parentFolders () {
+		// path of 'foo/bar/baz' would return folder objects whose paths are
+		// ['foo', 'foo/bar'] respectively
+		var parentFolders = [];
+		// iterate through parentFolder links to populate parentFolders
+		for (let i = this.parentFolder; i; i = i.parentFolder) {
+			parentFolders.push(i);
+		}
+		return parentFolders.reverse();
```
DavidVorick: there's no builtin `split` function?

Mingling94: This gets references to the folder objects, not the array of folder names (as the function `folderNames()` does)

plugins/Files/js/folder.js
```diff
@@ -105,26 +77,85 @@ var folder = Object.assign(Object.create(file), {
 		tools.waterfall(functs, names, callback);
 	},
 
-	// Share .sia files into folder at destination with same structure
+	// Share .sia files of all contents (deep) to destination
```
DavidVorick: this wording is not really clear

plugins/Files/js/folder.js
```diff
@@ -105,26 +77,85 @@ var folder = Object.assign(Object.create(file), {
...
+			qs: {
+				siapaths: this.paths,
+				destination: destination,
+			},
+		}, callback);
+	},
 
-		// Make folder at destination
-		mkdirp.sync(destination);
+	// Share ascii of all contents (deep)
```
DavidVorick: same thing. The word 'deep' really doesn't have any meaning

DavidVorick:
You gotta fix the build so it doesn't fail randomly.

Mingling94:
Oh nah this time it's because of linting errors that I neglected to check.

Mingling94:
@DavidVorick @lukechampine I think I've addressed all of #175 now...

plugins/Files/js/browser.js
```diff
@@ -226,34 +303,116 @@ var browser = {
...
+		if (itemCount === 0) {
+			tools.tooltip('No selected files', $('.controls .share').get(0));
+			return;
+		} else if (itemCount === 1) {
+			let file = files[0];
+			label = file.name;
+			paths = file.type === 'folder' ? file.paths : [file.path];
+		} else {
+			// Reduce array of files to array of all file paths
+			paths = files.reduce(function(a, b) {
```
lukechampine: reducing an array to an array sounds like a map, no? Or are the types different?

plugins/Files/js/browser.js
```diff
@@ -226,34 +303,116 @@ var browser = {
...
 			return;
 		} else if (itemCount === 1) {
+			// Save file/folder into specific place
 			label = files[0].name;
+			destination = tools.dialog('save', {
+				title: 'Download ' + label,
+				defaultPath: label,
+			});
 		} else {
+			let totalCount = files.reduce(function(a, b) {
```
lukechampine, line 401: woo higher order functions

plugins/Files/js/file.js
```diff
@@ -66,72 +83,86 @@ var file = {
 		}, callback);
 	},
 	share (destination, callback) {
-		if (typeof destination !== 'string' || destination.slice(-4) !== '.sia') {
-			console.error('Improper argument!', destination);
+		if (tools.notType(destination, 'string')) {
+			return;
+		} else if (destination.slice(-4) !== '.sia') {
+			console.error('Share path needs end in ".sia"!', destination);
```
lukechampine: typo, but this is handled in the backend anyway

plugins/Files/js/folder.js
```diff
@@ -48,103 +122,130 @@ var folder = Object.assign(Object.create(file), {
...
 
-		// Make corresponding array of file paths
-		names = names.map(name => `${destination}/${name}`);
+		var pathArray;
+		// File passed in is a file object
+		if (typeof f === 'object') {
+			pathArray = f.pathArray;
+		} else if (typeof f === 'string') {
+			// File passed in is a file path
+			pathArray = f.split('/');
```
lukechampine, line 191: I worry about the use of `/` as the path separator. Could break on Windows.

plugins/Files/js/fileElement.js
```diff
@@ -6,97 +6,115 @@
...
-	} else if (f.uploadProgress < 100) {
-		timeText = f.uploadProgress.toFixed(0) + '%'; 
+	// Set detail text
+	var detailText;
+	if (f.uploadprogress === 0) {
+		detailText = 'Processing...';
+	} else if (f.uploadprogress < 100) {
+		detailText = f.uploadprogress.toFixed(0) + '%'; 
+	} else if (f.renewing) {
+		detailText = 'Renewing';
```
lukechampine, line 35: `Renewing` makes it sound like something is actively happening, when in reality the file is just set to be renewed as needed. I think in this case, nothing should be displayed at all.

PR #184, opened 2016-01-15
Terminal plugin

irontiga:
Created a terminal plugin to connect to make it easy to talk to siac from within the wallet. Also had to add the fonts for Roboto Mono, without the terminal doesn't look like one. Requires a packaged version of Sia UI.


Mingling94:
Thanks! Screenshots?

The checks fail because your code didn't follow linting rules. Run `npm test` or `npm run lint` to see why and where it fails.

irontiga:
Here are some screenshots, I have also done a bit of a redesign, at least as much as I can do in a short amount of time.

irontiga:
I have no idea what is going on now. It's stuff about electron's main and render processes, stuff which I haven't messed with....

VoidingWarranties:
Can you divide this PR into multiple PRs, one for each logical group of changes? Specifically, separate the style changes to the UI from the terminal plugin. Thanks!

Mingling94:
@irontiga It's in the test script. The test is actually pretty trivial, check https://github.com/irontiga/Sia-UI/blob/master/test/main.js#L55 and change `(1, 6);` to `(1, 7);`

Sorry, I'm going to write some documentation on how these things work.

irontiga:
All good now?

Mingling94:
@irontiga They're hesitant to accept the changes because of @VoidingWarranties's point. It would be pretty easy to take _just_ the terminal plugin, but:
- the styling changes are controversial
- the update buttons aren't supposed to show when there's not an update.

Also @DavidVorick brought up a point that simply redirecting output to bash is way exploitable. If the plugin just executed one siac/siad command per line, it would be much safer. A malicious person could just be like "Oh yeah enter `siac -v && screwupyowholelife.sh`, it'll totally recover your wallet!"

DavidVorick:
Yeah, I definitely don't want users getting pwned because they don't know how to read bash and are blindly accepting commands from strangers. There's a limited amount that we can do to stop this, but at the very least we can make it so that 'siac' is the only program that gets run from the command line.

I believe we run 'siad' by starting a process and feeding it command line args through a struct. I think the sia.js node package has all the code explaining this.

In general, when working with open source it's best to make each separate thing its own PR. The terminal plugin should be one PR, and the style adjustments should be a different PR. That way each thing can be discussed separately.

pmknutsen:
@irontiga Would be real cool to see the Terminal plugin ready for next release.

irontiga:
Sorry, been unable to work on this for the last while. Will split it into 3 pr's, 1 for font updates, 1 for the plugin, and one for the ui's redesign.

DavidVorick:
Closing this PR because it has been a while. I still think the terminal plugin is a good idea, we will probably have official support for it later this summer.

PR #191, opened 2016-01-22
Restyling of Files plugin

pmknutsen:
Darkened text color in frame (file list)
Changed hover effects
Emphasize folders further with neon-green color
Dropdown menu and hover effects more consisent with rest of UI
Subtle gradient in plugin headers
Subtle vertical border on sidebar

screenshot from 2016-01-21 21 59 59
screenshot from 2016-01-21 21 58 28
screenshot from 2016-01-21 21 58 44

### DIFF
--- a/plugins/Files/css/files.css
+++ b/plugins/Files/css/files.css
@@ -8,6 +8,7 @@
  *		Grey:       #C5C5C5
  * 		Grey-Black: #4A4A4A
  * 		Black:      #000000
+ *      Neon-Green: #00CBA0
  */
 
 /* General */
@@ -41,9 +42,7 @@ input {
 	-webkit-animation: fadein 0.3s;
 }
 .directory:hover {
-	border: 1px solid #DDDDDD;
-	-webkit-box-shadow: inset 0 0 3px #ECECEC;
-	-webkit-animation: fadein 0.3s;
+	color: #00CBA0;
 }
 .button {
 	display: inline-block;
@@ -73,21 +72,39 @@ input {
 }
 .dropdown {
 	position: absolute;
-	z-index: 1;
-
-	border: 1px solid #dddddd;
-	border-radius: 0 6px 6px 6px;
-	margin: 0;
+	z-index: 2;
+	border: 1px solid #bdbdbd;
+	border-radius: 4px;
+	margin: 0 0 0 10px;
 	padding: 0;
-
 	white-space: nowrap;
 	background: white;
 	list-style: none;
+	overflow-x: auto;
+	width: 180px;
+
+
+	box-shadow: 0 0 16px 0 rgba(0, 0, 0, 0.25);
+
 }
 .dropdown .button {
 	display: block;
 	font-size: 12px;
-	margin: 4px;
+	padding: 6px 8px 6px 8px;
+	height: 25px;
+}
+.dropdown .button:hover {
+	background-color: #ECECEC;
+}
+.dropdown .button .icon {
+	width: 30px;
+	color: #00CBA0;
+}
+.dropdown .button .label {
+	font-size: 14px;
+	position: absolute;
+	left: 45px;
+	line-height: 25px;
 }
 hr {
 	margin: 0;
@@ -111,7 +128,7 @@ hr {
 }
 .file {
 	padding: 6px;
-	color: #757575;
+	color: #606060;
 	font-size: 14px;
 	background-color: #fff;
 	border-bottom: 1px solid #eeeeee;
@@ -121,8 +138,10 @@ hr {
 .file.selected {
 	background-color: #ddd;
 }
+.file:hover {
+	background-color: #ECECEC;
+}
 .file.anchor {
-	border: 1px solid #00CBA0;
 }
 .file:hover {
 	cursor: pointer;
@@ -173,7 +192,16 @@ hr {
 	text-shadow: 0 1px 0 #000;
 	white-space: nowrap;
 	z-index: 9999998;
+	font-size: 14px;
 }
 .file .fa-folder {
 	font-size: 16px;
+	color: #00CBA0;
+	padding-right: 3px;
+}
+.file .fa-file {
+	padding: 0 4px 0 1px;
+}
+.file .fa-refresh {
+	margin: 0 4px 0 1px;	
 }
--- a/plugins/Files/css/files.css
+++ b/plugins/Files/css/files.css
@@ -108,6 +108,8 @@ input {
 }
 hr {
 	margin: 0;
+	border: none;
+	border-bottom: 1px solid #eeeeee;
 }
 
 /* File list */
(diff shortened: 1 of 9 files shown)


DavidVorick:
While you are at it, can you find some alternative to the edit button? I'm guessing double-click doesn't work in the UI? The edit button looks pretty bad, especially next to every single file.

Maybe have a right-click menu on each file?

DavidVorick:
(thanks as always for contributing)

plugins/Files/css/files.css
```diff
@@ -121,8 +138,10 @@ hr {
 .file.selected {
 	background-color: #ddd;
 }
+.file:hover {
+	background-color: #ECECEC;
+}
 .file.anchor {
-	border: 1px solid #00CBA0;
```
Mingling94, line 125: Why get rid of the anchor selection box? I thought it was neat and made the effect of shift+click selections explicit. If you are getting rid of it, also delete lines 124 & 126 because they're unneeded clutter.

plugins/Files/index.html
```diff
@@ -30,41 +30,51 @@
 					</span>
 					<ul class='dropdown hidden'>
 						<li class='button'>
-							<span class='fa-stack'>
-								<i class='fa fa-folder-o fa-stack-2x'></i>
-								<i class='fa fa-plus fa-stack-1x'></i>
+							<span class="icon">
```
Mingling94, line 33: Nice! I suppose this aligns the icons ^_^

Mingling94:
@DavidVorick Originally the behavior was:
single-click the name -> edit name (This is how Windows does it)

Then @lukechampine suggested in #175 and I implemented:
single-click button -> edit name (Google Drive & Ubuntu-Gnome have it in context menu, but I haven't experimented with context menu options specific to the plugins yet. It may be non-trivial)
double-click files -> download (save dialog pops up)

Double-click has always worked, as evidenced by double-clicking folders to navigate inside them

Mingling94:
I think the green on the folders and dropdown selection are unnecessary. The green coloring so far is used for selections (as in the side bar) and emphasis (as in the capsule and for `button:hover` in the header of a plugin).

The green in the dropdown and folder icon are neither of these. Perhaps to emphasize the file/folder distinction, you could add another column, `Type`, like how Ubuntu-Gnome does it:
selection_007

DavidVorick:
Or maybe a sia-purple could be used? I like the idea of coloring the folders differently, but it sounds like green was the wrong choice.

pmknutsen:
I agree the color choice was poor; the neon green should be reserved for selected or high-lighted items. In case you didn't notice, I also used the green when hovering over directory names (see screenshot below; note pointer isn't seen).

I removed the folder color and added a Type column, which I think is a good idea.

screenshot from 2016-01-22 08 56 00

screenshot from 2016-01-22 08 54 42

pmknutsen:
This how it would look with purple. I think it clashes a bit with the rest.

screenshot from 2016-01-22 09 01 30

lukechampine:
I think the current combination of 1) slightly larger icon, 2) `/` after name, and 3) type "Folder" is enough to make folders distinguishable from files. A separate color is a sensible approach to try, but it seems difficult to pick the proper color without it looking gaudy. I think it's ok for now.

DavidVorick:
Are you only seeing 17 active hosts?

Jinglan:
I think that one helpful differentiation btwn folders and files would be this:
archive142
text70

Folder icon is dark filled, file icon is white filled.

DavidVorick:
that looks good to me

pmknutsen:
@Jinglan Agreed. That makes it clearer.

screenshot from 2016-01-22 11 03 28

lukechampine:
excellent!

DavidVorick:
I feel like the 'type' column is not needed, especially with the dark/light distinction between folders and files. I think the change is good.

I still cringe when I look at the edit-pencils. That doesn't need to be fixed in this PR though.

lukechampine:
agreed. We can consider readding "type" when it would add more information, like in Ming's screenshot above (though this may not be currently possible; .sia files don't contain that metadata).
The edit-pencils should be moved to a single button, alongside the others at the top.
btw, there's a bug related to renaming folders: siad doesn't support renaming folders, just files. So really when you rename a folder, the UI should actually be renaming all the files within that folder. Alternatively, we could make this siad's job.
Supporting drag-and-drop to move files in and out of folders would also be a great feature to have.

pmknutsen:
@lukechampine I removed Type column. Context menu won't be a problem for renaming a file etc, but there is a conflict with the window context-menu when right-clicking. @Mingling94 How to prevent the global context menu from showing when a specific element is clicked?

lukechampine:
It may make more sense to hold off on right-clicking behavior for now. If we want to add it, I think we'd also have to add support for all the usual actions (delete/download/rename) as well as right-clicking with multiple files selected.
on that subject, I think if/when we add right-click actions, they shouldn't replace the actions already visible in the top bar. The user may not be aware that the right-click menu exists.

DavidVorick:
siad should be the one that supports renaming folders.

Mingling94:
@lukechampine
- Renaming a folder in the UI does just rename all the files within said folder. I've checked that this feature works with at least one file like: `Folder/File` -> `RenamedFolder/File`
- Also a rename button wouldn't really make sense as a top bar button because it doesn't apply to many items.
- Drag and drop would be cool

@pmknutsen 
- I wouldn't know exactly. Check out: https://github.com/NebulousLabs/Sia-UI/blob/master/index.html#L60 Since an event is what fires the CM, and it clearly bubbles up from webviews to the general index.html, you'd have to change these lines to only fire a global CM if the webview itself doesn't have a more specific one. Though like luke said, it can be saved for another PR.

lukechampine:
Whoops, correct you are. Not sure why I was mistaken about that. Maybe I was thinking of renaming empty directories, since siad doesn't understand them.

Good point on the rename button. It could simply be grayed out when multiple files were selected though.
Looking at other directory interfaces, they seem to get by without a rename button just fine. I think at this point the user expects two methods for renaming:
- Right click and select "Rename"
- Click once to highlight, wait for double-click timeout, then click again

Also, most UIs will highlight the filename when renaming. Might be a nice touch.

pmknutsen:
Current version after commit 48cfd0c looks like this:

screenshot from 2016-01-23 15 51 01

@lukechampine The file row remains highlighted when renaming. Auto-selecting the entire filename would be nice too.

Rename icon simply does nothing if multiple files are selected. Grayed out or not showing at all would be better.

lukechampine:
nice. Looks like the build failed due to some frivolous syntax errors. Once those are cleaned up, I think this should be merged. Further changes can be made in a new PR.

pmknutsen:
@lukechampine Fixed the syntax errors but Travis build still fails. I don't understand the error. Could you have a look plz?

lukechampine:
just some nondeterminism. I've rerun the build. Thanks!

PR #192, opened 2016-01-22
More Files plugin fixes

lukechampine:
Single-file downloads were not being downloaded to the correct location, since the download destination was constructed as: `${destination}/${file.name}`. For multiple files this makes sense, but for a single file it
constructs the wrong path; we just want `${destination}`.

Also replaced one use of `/` with `path.join`, to create proper paths on Windows. There may be some other uses of `/` that cause problems, but it seems unlikely; node apparently interprets `/` as the path separator regardless of operating system.

### DIFF
--- a/plugins/Files/js/browser.js
+++ b/plugins/Files/js/browser.js
@@ -400,7 +400,10 @@ var browser = {
 			tools.tooltip('No selected files', $('.controls .download').get(0));
 			return;
 		} else if (itemCount === 1) {
-			// Save file/folder into specific place
+			if (files[0].type === 'folder') {
+				tools.notify('No support for downloading folders (yet)', 'error');
+				return;
+			}
 			label = files[0].name;
 			destination = tools.dialog('save', {
 				title: 'Download ' + label,
@@ -414,15 +417,14 @@ var browser = {
 				tools.notify(`Downloaded ${label} to ${destination}`, 'success');
 			});
 		} else {
-			let totalCount = files.reduce(function(a, b) {
-				if (b.type === 'folder') {
-					return a + b.count;
-				} else {
-					return ++a;
+			for (let file of files) {
+				if (file.type === 'folder') {
+					tools.notify('No support for downloading folders (yet)', 'error');
+					return;
 				}
-			}, 0);
-			// Save files/folders into directory
-			label = totalCount + ' files';
+			}
+			// Save files into directory
+			label = itemCount + ' files';
 			destination = tools.dialog('open', {
 				title: 'Download ' + label,
 				properties: ['openDirectory', 'createDirectory'],
@@ -432,11 +434,11 @@ var browser = {
 			}
 			// Download each of the files
 			tools.notify(`Downloading ${label} to ${destination}`, 'download');
-				let functs = files.map(file => file.download.bind(file));
-				let destinations = files.map(file => path.join(destination, file.name));
-				tools.waterfall(functs, destinations, function() {
-					tools.notify(`Downloaded ${label} to ${destination}`, 'success');
-				});
+			let functs = files.map(file => file.download.bind(file));
+			let destinations = files.map(file => path.join(destination, file.name));
+			tools.waterfall(functs, destinations, function() {
+				tools.notify(`Downloaded ${label} to ${destination}`, 'success');
+			});
 		}
 	},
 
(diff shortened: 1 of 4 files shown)


plugins/Files/js/browser.js
```diff
@@ -420,20 +427,17 @@ var browser = {
 				title: 'Download ' + label,
 				properties: ['openDirectory', 'createDirectory'],
 			});
+			if (!destination) {
+				return;
+			}
+			// Download each of the files
+			tools.notify(`Downloading ${label} to ${destination}`, 'download');
+				let functs = files.map(file => file.download.bind(file));
```
Mingling94: Tabbing, might as well fix it.

DavidVorick:
I am in favor of using `/` everywhere, even if node reinterprets it correctly on Windows. Experience tells me that being pedantic pays off when reliability is a goal.

DavidVorick:
Pull request looks good to me, but the tabbing should be fixed first.

lukechampine:
I don't understand. I think you made a typo.

Anyway, we probably shouldn't use `path.join` everywhere, since it means Windows users will create siapaths with `\` and others will create siapaths with `/`, and I don't know if that will cause incompatibility when sharing .sia files. It might be best to use `/` everywhere and convert the Windows `\` as soon as it's passed in.

Wait, hmm...if we use `/` internally, will it break siac on Windows? Calling `siac renter delete foo/bar/baz` on Windows might cause a parsing error. Ugh.

DavidVorick:
Yeah I meant to say `path.Join` everywhere.

Urg and what happens when you share files across filesystems? Internally, `Sia` should definitely use `/` and then switch to the compatible version when going to disk.

hmph

plugins/Files/js/browser.js
```diff
@@ -61,6 +61,7 @@ function getSelectedFiles() {
 	var selected = $('.selected.file');
 	var nameFields = selected.map((i, el) => $(el).find('.name')).get();
 	var names = nameFields.map(field => field.text());
+	console.log(names);
```
DavidVorick: did you mean to leave this here?

lukechampine:
at any rate, this PR is good to go. We can address these things more thoroughly in 0.5.1

plugins/Files/js/browser.js
lukechampine: damn

plugins/Files/js/browser.js
```diff
@@ -375,7 +376,7 @@ var browser = {
 			// Place siafile to location
 			tools.notify(`Sharing ${label} to ${destination}`, 'siafile');
 			loader.shareDotSia(paths, destination, function() {
-				tools.notify(`Put ${label}'s .sia files at ${destination}`, 'success');
+				tools.notify(`Put ${label}s .sia files at ${destination}`, 'success');
```
DavidVorick: is it correct to be removing the apostrophy? What sorts of strings does ${label} expand to?

lukechampine: just lazy committing on my part.
I removed the apostrophe while coding because it messed up my syntax highlighting

plugins/Files/js/global.js
```diff
@@ -85,7 +85,7 @@ $('.dropdown .button').click(function() {
 			$('.dropdown li').hide('fast');
 			$('#paste-ascii').show('fast');
 			$('#paste-ascii input').focus();
-			return; // Don't close dropdown
+			break;
```
Mingling94, line 88: The return here was needed to not close the dropdown.

plugins/Files/js/global.js
```diff
@@ -94,20 +94,17 @@ $('.dropdown .button').click(function() {
 			break;
 		default:
 			console.error('Unknown button!', this);
-			return; // Don't close dropdown
-	}
-
-	// Detect flawed userInput from actions that require it, hide if fine
-	if (!userInput && option !== 'Make Folder') {
-		tools.tooltip('Invalid action!', this);
```
Mingling94, line 102: If no input, silently fail and close the dropdown? Is that better?

plugins/Files/js/global.js
lukechampine, line 88: whoops :I
thanks for popping in to help us debug. We have our hands full at the moment.

PR #193, opened 2016-01-24
File type specific icons

pmknutsen:
Display icons specific to the file extension.

screenshot from 2016-01-23 22 12 18

### DIFF
--- a/plugins/Files/js/fileElement.js
+++ b/plugins/Files/js/fileElement.js
@@ -9,6 +9,25 @@
 const $ = require('jquery');
 const tools = require('./uiTools');
 
+// File icons by extension
+const fileIcons = {
+	'pdf':'file-pdf-o',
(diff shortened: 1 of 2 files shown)


DavidVorick:
This will probably make the UI feel a lot more comfortable overall. Thanks, I'm glad you did this.

plugins/Files/js/fileElement.js
```diff
@@ -47,14 +77,22 @@ function updateFileElement(f, el) {
 
 // Make file element with jquery
 function makeFileElement(f) {
+	// Get extension from name
+	var re = /(?:\.([^.]+))?$/;
+	var ext = re.exec(f.name)[1];
```
Mingling94, line 82: Alternatively, you could use `path.extname()` https://nodejs.org/api/path.html#path_path_extname_p

PR #194, opened 2016-01-24
Remove screen header

pmknutsen:
This proposal is inspired by @irontiga's PR that removes the screen header and moves the update icons to the sidebar.

Changes:
- Remove screen header
- Move update icons to sidebar (show only when there is an update)

screenshot from 2016-01-23 23 39 53
screenshot from 2016-01-23 23 39 26

### DIFF
--- a/css/general.css
+++ b/css/general.css
@@ -40,22 +40,19 @@ html,body {
 }
 .body {
 	position: relative;
-
-	height: -webkit-calc(100% - 80px);
-	-webkit-box-sizing:border-box;
+	height: 100%;
+	-webkit-box-sizing: border-box;
 }
 
 /* UI Header */
 .header {
-	height: 80px;
+	height: 60px;
 	width: 100%;
-
-	background-color: #313133;
-	background: linear-gradient(to left, #313133 15%, #0b0e16 85%);
+	background-color: #0b0e16;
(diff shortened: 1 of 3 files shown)


lukechampine:
Update buttons in the sidebar feel a bit... overbearing.
Maybe we could move them inside the About plugin? Or only display them as a startup notification.
Also, `Update Sia` hasn't been functioning for a while now, so it can be removed entirely.

By the way, @Jinglan is currently designing a new look for the UI. Not sure how drastic her changes will be, but it might be best to hold off on further layout changes -- it'd be a shame to have to throw them away if they conflict with her designs. I'd like her to weigh in on this.

DavidVorick:
I don't really like this change.

pmknutsen:
In any case, for future consideration I think the top header needs to go if it will remain empty. Should I close this PR?

DavidVorick:
waiting for comments from Jing, our designer.

Jinglan:
I think better to have "notifications" between "wallet" and "about" on the sidebar, so you can click on "notifications" and see "New notification: UI update pending" or something like that.

lukechampine:
While we aren't going to merge this outright, I think this layout may influence our future design.

PR #195, opened 2016-01-25
Wallet plugin fixes

pmknutsen:
Fixed broken Address search
More intuitive page navigation
Enforce limits on requested page
Compare _absolute_ transaction values in transaction filter
Add a reasonable minimum value to avoid displaying dust
Increase font size in Wallet to 12px
Darker red and green on transaction icons
Improved formatting of small Siacoin amounts (show n'th significant digits vs 2 decimals for amounts > 1 SC)

screenshot from 2016-01-24 22 30 00

### DIFF
--- a/plugins/Wallet/js/helpers.js
+++ b/plugins/Wallet/js/helpers.js
@@ -13,8 +13,8 @@ function convertSiacoin(hastings) {
 	var number = new BigNumber(hastings);
 	var ConversionFactor = new BigNumber(10).pow(24);
 
-	return number.gt(1) ? number.dividedBy(ConversionFactor).round(2)
-						 : number.dividedBy(ConversionFactor).toPrecision(1);
+	return number.gt(1) ? number.dividedBy(ConversionFactor).round(2).toNumber()
+						: number.dividedBy(ConversionFactor).toPrecision(1);
 }
 
 // Notification shortcut
--- a/plugins/Wallet/js/helpers.js
+++ b/plugins/Wallet/js/helpers.js
@@ -12,10 +12,12 @@ function convertSiacoin(hastings) {
 	// This deals with that imperfectly
 	var number = new BigNumber(hastings);
 	var ConversionFactor = new BigNumber(10).pow(24);
-	return number.dividedBy(ConversionFactor).round(2);
+
+	return number.gt(1) ? number.dividedBy(ConversionFactor).round(2)
+						 : number.dividedBy(ConversionFactor).toPrecision(1);
 }
 
-// Notification shortcut 
+// Notification shortcut
 function notify(msg, type) {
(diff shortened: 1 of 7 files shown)


plugins/Wallet/js/transactions.js
```diff
@@ -33,8 +33,8 @@ function checkCriteria(txn) {
 	// Failing one check returns false but passing all checks returns true
 	return (criteria.endTime   === null || txn.confirmationtimestamp <= criteria.endTime)   &&
 	       (criteria.startTime === null || txn.confirmationtimestamp >= criteria.startTime) &&
-	       (criteria.maxValue  === null || txn.value                 <= criteria.maxValue)  &&
-	       (criteria.minValue  === null || txn.value                 >= criteria.minValue)  &&
+	       (criteria.maxValue  === null || Math.abs(txn.value)       <= criteria.maxValue)  &&
+	       (criteria.minValue  === null || Math.abs(txn.value)       >= criteria.minValue)  &&
```
Mingling94, line 37: Great catch!

plugins/Wallet/js/transactions.js
```diff
@@ -90,7 +90,7 @@ function makeTransaction(txn) {
 	element.find('.time').html(time);
 
 	// Set transaction positive or negative
-	if (txn.value.isNegative()) {
+	if (txn.value < 0) {
```
Mingling94, line 93: Does this work? I thought from line 68, `txn.value` is a `BigNumber` object?

pmknutsen, line 93: It works. For some reason I don't understand, when you call toPrecision() on the BigNum it seems to be cast to a real number and isNegative() is no longer an available method. I tested it on negative and positive cases.

DavidVorick:
Looks good to me.

plugins/Wallet/js/helpers.js
```diff
@@ -12,10 +12,12 @@ function convertSiacoin(hastings) {
 	// This deals with that imperfectly
 	var number = new BigNumber(hastings);
 	var ConversionFactor = new BigNumber(10).pow(24);
-	return number.dividedBy(ConversionFactor).round(2);
+
+	return number.gt(1) ? number.dividedBy(ConversionFactor).round(2)
+						 : number.dividedBy(ConversionFactor).toPrecision(1);
```
Mingling94: Ah then this is the problem. `round()` returns a `BigNumber` and `toPrecision()` returns a `string`. This function should return one type of value.

plugins/Wallet/js/transactions.js
Mingling94, line 37: These lines also depend on what type of value is returned from `convertSiacoin()` and set per transaction object.

PR #208, opened 2016-02-16
Don't start plugins until after loading screen

Mingling94:


index.html
```diff
@@ -61,7 +61,7 @@
 			// It initiates all other functionality, directly or indirectly
 			var UI = require('./js/rendererjs/uiManager.js');
 			// Need this for devtools to work
-			var Plugins = require('./js/rendererjs/pluginManager.js');
+			var Plugins;
```
lukechampine: where does this value eventually get set? I see `plugins` and `ui` being referenced in the other files; don't they need to be lower-cased here as well?

Mingling94: Oops, forgot to correct the index.html. That's what happens when you only grep inside a js folder, this would have broke all the things.

Issue #205, opened 2016-02-06
Siad needs to be running


JoshuaMurphynz:
when i run

Sia-UI

I get an error saying that siad needs to be running.

This should be started automatically

lukechampine:
siad is started automatically, though it can take a while to startup. Personally, I prefer to run siad separately from the UI.

DavidVorick:
I do think that the UI should not close 'siad'. People were originally confused by the behavior that 'siad' would not shut down when the UI was closed, but it's important to the hostdb that 'siad' be running most of the time otherwise it won't know which hosts have good uptime.

avahowell:
This behavior has been specified and implemented as of #220 and #222.  I am therefore closing this issue.

PR #210, opened 2016-02-16
By default, siad is detached and outlives the UI

Mingling94:
Mentioned as desired behavior in #205

### DIFF
--- a/js/mainjs/config.js
+++ b/js/mainjs/config.js
@@ -5,7 +5,7 @@ const defaultConfig = {
 	homePlugin:  'Overview',
 	siad: {
     	path: require('path').join(__dirname, '../..', 'Sia'),
-		detached: false,
+		detached: true,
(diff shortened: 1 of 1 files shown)


lukechampine:
I'm not sure about this, as it means siad has to be closed via siac, which most newbies aren't aware of.
Maybe best would be a confirmation dialog when closing that asks if you want to shut down siad as well?

pmknutsen:
I think the desired behavior is that closing the window, will simply minimize Sia-UI.

Sia-UI should be shut down with either Ctrl+Q, selecting Quit in a window menu, or a Quit option in a toolbar menu. This is how most UIs that want to be persistent do it.

Quit'ing Sia-UI should then also shutdown siad.

Mingling94:
Ah, good to restart the conversation about desired behavior then!

PR #211, opened 2016-02-16
On wrong password, show popup and delete saved pw

### LINKED ISSUE #200: Passphrase saving even if wrong
Yosemite Darwin 0.5.1

Saving and sending siad the password and unlock command seems to work fine, but there is no going back if you put in the wrong password. If you fat finger it, you don't get the chance to change the string and you are condemned to just the get Wrong Password error.

Fix: Make Sia forget the password if you enter the wrong one
       or
       Make Sia save the password only if it's the correct one

Mingling94:
Fixes #200

### DIFF
--- a/plugins/Wallet/.jshintrc
+++ b/plugins/Wallet/.jshintrc
@@ -120,6 +120,8 @@
 		"tooltip": true,
 		"encrypt": true,
 		"lock": true,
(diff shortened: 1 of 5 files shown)


lukechampine:
thanks!

Issue #149, opened 2015-12-14
Oversized Sia logo in OS X's status menu.


VoidingWarranties:
![screen shot](https://cloud.githubusercontent.com/assets/5420909/11786925/46e2e920-a257-11e5-9484-f0ee62efdf32.png)

DavidVorick:
I blame Apple, the OS should be able to resize as necessary. I'm pretty sure that's what both Linux and Windows does.

heh. We should fix it.

VoidingWarranties:
The icon shouldn't even exist in the status menu unless there are clickable items or other display info in it. Are there actions that the status menu currently provides? If there are none we should remove it.

DavidVorick:
I think it was supposed to have the buttons that allowed you to minimize the window or something.

I'm not sure how it got there, you'd have to talk to Ming.

@Mingling94 ideas?

Issue #148, opened 2015-12-14
OS X specific actions don't work (quit, hide, etc...)


VoidingWarranties:
e.g. command-Q for quit, command-H for hide. There are others as well, such as command-option-H for hide others. There also aren't menu bar items for these actions. The only way to exit the program in is to click the red x, and there is no way to hide the program AFAIK.

lukechampine:
fixed in #212

PR #212, opened 2016-02-17
Fixes for #149 and #148

avahowell:


js/mainjs/trayMenu.js
```diff
@@ -0,0 +1,22 @@
...
+	// Template for Sia-UI tray menu.
+	var menutemplate = [
+		{
+			label: 'Show Sia',
+			click: function () { window.show(); }
+		},
+		{ type: 'separator' },
+		{
+			label: 'Hide Sia',
+			click: function () { window.hide(); }
```
lukechampine, line 16: could these just be `window.show` and `window.hide`?
although, even if it's possible, I'm not sure it's best practice.

VoidingWarranties:
Can you close the issues in the commit messages both so that they are closed automatically and so that it's obvious what each commit fixes

js/mainjs/initWindow.js
```diff
@@ -49,13 +48,12 @@ module.exports = function(config) {
...
 	// Load the index.html of the app.
 	mainWindow.loadURL('file://' + __dirname + '/../../index.html');
-
 	// Choose not to show the menubar
 	if (process.platform !== 'darwin') {
 		mainWindow.setMenuBarVisibility(false);
 	} else {
 		// Create the Application's main menu - OSX version might feel weird without a menubar
-		Menu.setApplicationMenu(AppMenu);
+		Menu.setApplicationMenu(new AppMenu(mainWindow));
```
lukechampine: I still haven't fully grasped the `new` operator. What purpose does it serve here?

index.js
```diff
@@ -22,13 +23,14 @@ var config = require('./js/mainjs/config.js')(Path.join(__dirname, 'config.json'
 
 // When Electron loading has finished, start the daemon then the UI
 App.on('ready', function() {
+	// Load mainWindow
+	mainWindow = require('./js/mainjs/initWindow.js')(config);
 	// Load tray icon
-	var iconPath = Path.join(__dirname, 'assets', 'icon.png');
+	// OSX has a maximum tray icon height of 18px.
```
lukechampine: can probably drop this comment. It's more confusing than helpful.

VoidingWarranties: Where does 18px come from? http://iconhandbook.co.uk/reference/chart/osx/ says the maximum size is 22px. Also is it possible to use the previous larger icon (or vector image if we have that), but display it at a smaller size?

VoidingWarranties:
Where does 18px come from for the OS X menu bar size? http://iconhandbook.co.uk/reference/chart/osx/ says the maximum size is 22px. Also is it possible to use the previous larger icon (or vector image if we have that), but display it at a smaller size?

js/mainjs/initWindow.js
avahowell: I amended the commits to reflect their issue fixes.  I initially called this as Menu.setApplicationMenu(AppMenu()), and JSHint yelled at me for not calling a constructor using new.  Looking into it more, it looks like jshint expects all functions that start with caps to be constructors, invoked using new.  AppMenu() can still be called as a function, but should probably be renamed to reflect this convention

lukechampine: gotcha. I agree that `AppMenu` should be lowercased.

avahowell:
Electron's nativeImage class only supports png and jpeg, unfortunately no vector images.  I chose 18px to give some vertical margin to the icon, the maximum tray icon size is indeed 22px on osx.

Issue #213, opened 2016-02-17
Proposal for handling UI exit


pmknutsen:
![ui_close_proposal](https://cloud.githubusercontent.com/assets/1565499/13099654/4f6cf396-d4e9-11e5-97d1-bd17c45fe65c.png)

The above is just a mockup (ignore that I'm using 0.4.8 as backdrop)

This is a proposal for how to deal with user either wanting to just exit the UI, or shut down Sia altogether.

I removed the window frame (should also add a border)
An X was put back in the upper right hand corner
A Quit button was added on the Overview page

The intended behavior is:
Pressing the X minimizes the UI
Pressing "Quit" will exit the UI _and_ `siad`

Selecting "Quit" should probably also launch a dialog warning user that the host will be shut down as well, and then provide a Confirm button.

lukechampine:
@johnathanhowell is planning on implementing the approach you described in #210, where closing minimizes to tray and the tray contains a "Quit" option. I think this is a good approach, in line with how many apps work when they need to run in the background.

VoidingWarranties:
For desired behavior when quitting the UI, what about this?
- If there is already an instance of `siad` running (on the default port), do not quit `siad` when quitting the UI.
- If no instance of `siad` is running when the UI starts, the UI starts one and quits it when the UI quits.

This was the desired behavior @DavidVorick mentioned a few weeks ago when we were discussing it offline.

pmknutsen:
@VoidingWarranties A remaining problem is users that run a host, started when the UI is opened. When the UI is closed, the host shuts down too in your scheme.

Mingling94:
@VoidingWarranties There is already an option in the siad wrapper and defaulted to `false` in the UI called `detached`.
if `true` - run siad as either a detached, background process and doesn't exit it with the UI's closing.
if `false` - start siad if not already running, close upon the UI's closing.

I didn't make a settings page/plugin ( #47  ) to change said setting, but if you changed it in the config.json manually, you'd see the results.

Mingling94:
I think this suggestion of providing buttons to close siad with the UI is approaching the problem from a more action-based perspective, but I'd sooner like to see a settings plugin where one can change the persistence of siad.

This means that hosts could set `detached` to `true` and renters/whoever else could set/leave it as `false`

VoidingWarranties:
@pmknutsen hmmm good point. I like your suggested approach. Then again maybe the host is somewhat of a special case and we could implement something like:
[...]

PR #214, opened 2016-02-17
UI exit handling from the tray

avahowell:
The mainWindow.close event is now overridden and mainWindow.minimize() is called instead.  The only way to quit the application is through the tray menu, all other calls to mainWindow.close() (cmd-Q, the x button, etc) minimize the main window instead.  There is not currently a confirmation dialog.  The previous behavior with config.siad.detached remains intact: quitting the application will not kill siad if this flag is set.

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,14 +17,16 @@ const Path = require('path');
 // garbage collection
 var mainWindow;
 var appIcon;
+var siad;
 
 // config.json manager
 var config = require('./js/mainjs/config.js')(Path.join(__dirname, 'config.json'));
 
 // When Electron loading has finished, start the daemon then the UI
-App.on('ready', function() {
+App.on('ready', function () {
 	// Load mainWindow
 	mainWindow = require('./js/mainjs/initWindow.js')(config);
+
 	// Load tray icon and menu
 	var iconPath = Path.join(__dirname, 'assets', 'tray.png');
 	appIcon = new Tray(iconPath);
@@ -34,31 +36,17 @@ App.on('ready', function() {
 	// Add IPCMain listeners
 	require('./js/mainjs/addIPCListeners.js')(config, mainWindow);
 
-	// If config.persistInTray is set, catch mainWindow's close event and minimize instead.
-	if (config.persistInTray) {
-		mainWindow.on('close', function(e) {
-			if (!mainWindow.wantsQuit) {
-				e.preventDefault();
-				mainWindow.minimize();
-			}
-		});
-	}
 	// Load siad
-	var Siad = require('./js/mainjs/initSiad.js')(config, mainWindow);
-	// mainwindow.closed() can only be called when the user quits from the tray.
-	// If the config.siad.detached flag is not set, call siad.stop and quit from the callback.
-	mainWindow.on('closed', function() {
-		// save the config on exit
-		config.save();
-		if (!config.siad.detached) {
-			Siad.stop(function() {
-				App.quit();
-			});
-		} else {
-			App.quit();
-		}
-		// Dereference mainWindow and Siad to clean up.
-		Siad = null;
-		mainWindow = null;
-	});
+	siad = require('./js/mainjs/initSiad.js')(config, mainWindow);
+});
+// Quit once all windows have been closed.
+// If the config.siad.detached flag is not set, call siad.stop and quit from the callback.
+App.on('window-all-closed', function () {
+	config.save();
+	if (!config.siad.detached) {
+		siad.stop();
+	} 
+	App.quit();
+	mainWindow = null;
+	siad = null;
 });
--- a/index.js
+++ b/index.js
@@ -25,25 +25,25 @@ var config = require('./js/mainjs/config.js')(Path.join(__dirname, 'config.json'
 App.on('ready', function() {
 	// Load mainWindow
 	mainWindow = require('./js/mainjs/initWindow.js')(config);
-	// Load tray icon
+	// Load siad
+	require('./js/mainjs/initSiad.js')(config, mainWindow);
+	// Load tray icon and menu
 	var iconPath = Path.join(__dirname, 'assets', 'tray.png');
 	appIcon = new Tray(iconPath);
 	appIcon.setToolTip('Sia - The Collaborative Cloud.');
-	appIcon.setContextMenu(appTray(mainWindow));
+	appIcon.setContextMenu(appTray(mainWindow, App));
 
 	// Add IPCMain listeners
 	require('./js/mainjs/addIPCListeners.js')(config, mainWindow);
 
-	// Upon exiting, dereference the window object so that the GC cleans up.
-	mainWindow.on('close', function() {
-		mainWindow = null;
+	// Catch mainWindow's close event and minimize instead.
+	mainWindow.on('close', function(e) {
+		e.preventDefault();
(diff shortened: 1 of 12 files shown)


VoidingWarranties:
It seems sneaky to override `mainWindow.close`. To preserve semantics, perhaps it would be better to change all the calls to `mainWindow.close` to `mainWindow.minimize` instead, and have the quit button call `mainWindow.close`. What do you think?

avahowell:
I agree it would be better not to override `mainWindow.close`.  I couldn't find a way to change `BrowserWindow`'s window controls to call `minimize` instead of `close`.  Perhaps it would be better to instead use `window.beforeunload` as described here: https://github.com/atom/electron/blob/master/docs/api/browser-window.md#event-close.  Maybe add a flag to `mainWindow`, `wantsQuit`, that is set by the tray menu's quit button.

There are larger issues to discuss here as well.  For example, what if the system has no notification tray?  The user will be unable to close Sia-UI.  Because of this we probably want a `Quit` button somewhere in the UI as well.

pmknutsen:
@johnathanhowell What about removing the window frame altogether (along with min/max/close) buttons, and insert a close button in HTML plus Quit button in UI. Then there is no reason to override `mainWindow.close` etc.  See proposal #213. This way, all is contained within the UI, and no per-OS toolbar manipulations will be required.

avahowell:
I think @pmknutsen's proposal is a good solution to the issues that have been brought up thus far.  I'm implementing it now.

pmknutsen:
For some ideas, see http://www.mylifeforthecode.com/making-the-electron-shell-as-pretty-as-the-visual-studio-shell/

VoidingWarranties:
I'm worried that clicking the "x" and seeing the window be minimized isn't what the user would expect. The way other applications handle this UX (at least on OS X) is to have the main window be closeable, but able to run without a main window. In other words, when the user clicks the x, the window dissapears but the application continues running and still appears in the system tray and taskbar / dock. The window appears again when the user clicks the app's icon either in the system tray, or taskbar / dock. This seems like a bigger change for another PR though.

For this PR, I would also hesitate to replace the OS chrome with our own. The standard location for close / minimize buttons on Windows is on the right hand side of the window, but on OS X it's on the left. Is it possible to disable the OS close button but leave the OS minimize button enabled? I know this is possible for OS X and Windows natively, but I'm not sure if electron supports it. That way there's no need to override `mainWindow.close`, and no need to change any of the calls to `mainWindow.close` to `mainWindow.minimize`.

avahowell:
It is, but only on Windows and OSX.  According to the electron docs, 
`minimizable Boolean - Whether window is minimizable. This is not implemented on Linux. Default is true.`
`maximizable Boolean - Whether window is maximizable. This is not implemented on Linux. Default is true.`
`closable Boolean - Whether window is closable. This is not implemented on Linux. Default is true.`

I think these not being implemented on Linux is probably a dealbreaker for us.  As for the "x", we could have it call `mainWindow.hide()` instead of `.minimize()`.  I think that's more consistent with the behavior expected of "x", but we're back to the problem of requiring a working tray to use Sia-UI effectively.

VoidingWarranties:
The tray doesn't need to be fully functional. If clicking the tray unhides the window that would be sufficient.

lukechampine:
can we detect the lack of a system tray? If so, shutting down completely seems like reasonable fallback behavior. I think the number of users (UI users, at that!) without a system tray will likely be a small minority.
also, worth nothing that most apps with this behavior provide a "close to tray?" checkbox somewhere. I don't think it's strictly necessary, but might be nice to have down the line.

pmknutsen:
Also, what happens currently when a second instance of Sia-UI is launched? One way to recover from a vanished UI is to run Sia-UI again but seeing there is already a running instance then pop that UI to front. As a fail-safe for worst-case scenarios.

js/mainjs/trayMenu.js
```diff
@@ -8,15 +8,22 @@ module.exports = function (window) {
...
 			label: 'Hide Sia',
-			click: function () { window.hide(); }
+			click: function () { window.minimize(); }
 		},
+		{ type: 'separator' },
+		{
+			label: 'Quit Sia',
+			click: function() {
+				window.wantsQuit = true;
+				window.close();
```
VoidingWarranties: What was wrong with `window.destroy()`? Adding additional state seems uglier to me. But maybe it's JS convention, idk.

avahowell:
@pmknutsen currently, if you minimize one instance of Sia-UI and launch another, the other instance will launch without issue and both will communicate to the same `siad`.  They are still two separate UI instances however.  

@lukechampine The electron docs do not mention any functionality to detect an accessible system tray.  As for the checkbox, maybe that's something for the Settings plugin?  I can throw it in the config for now.  Something like `persistInTray`?

As of now I'm going to leave the earlier suggestions about a frameless UI out of this PR.  I replaced the .destroy() call in trayMenu with a close() call, using a flag `wantsQuit`.

js/mainjs/trayMenu.js
avahowell: According to the docs, destroy does this:

`Force closing the window, the unload and beforeunload event won't be emitted for the web page, and close event will also not be emitted for this window, but it guarantees the closed event will be emitted.`

This means any event listeners assigned to unload, beforeunload, and close() wont fire.  There's two listeners for these events in Sia-UI,
`window.onbeforeunload = closeLog` .
in uiManager.js at L155 and
`mainWindow.on('close'...` 
in initWindow.js at L36

Neither of these get called if you just call `.destroy()`.

VoidingWarranties: What about creating a custom event to handle quitting? Something like `window.quit`. Or since we want to have all of the other events fire when quitting, create a custom event for "closing" and have `window.close` do the "quitting".

avahowell:
Here's where this is at now:  by default, all BrowserWindow.close() calls hide the main window.  The user can use the tray menu to bring the window back, or to quit the application.  This is the default behavior, it's toggleable via the persistInTray boolean in config.  BrowserWindow.close() is no longer overridden using e.preventDefault, instead I used window.onbeforeunload to hide the main window and cancel the close() event.  The main window, and therefore application, can be gracefully quit by calling mainWindow.destroy().

js/mainjs/config.js
```diff
@@ -1,12 +1,13 @@
...
 
-// The default settings 
+// The default settings
 const defaultConfig = {
 	homePlugin:  'Overview',
 	siad: {
     	path: require('path').join(__dirname, '../..', 'Sia'),
 		detached: false,
 	},
+	persistInTray: true,
```
lukechampine: can we call this `closeToTray` instead?
Also, I'm thinking that since linux users may not have a tray, this should be `false` by default for linux users:

``` js
// Only Windows and OS X are guaranteed to have a system tray
closeToTray: process.platform === 'win32' || process.platform === 'darwin',
```

index.js
```diff
@@ -22,10 +22,11 @@ var appIcon;
 var config = require('./js/mainjs/config.js')(Path.join(__dirname, 'config.json'));
 
 // When Electron loading has finished, start the daemon then the UI
-App.on('ready', function() {
+App.on('ready', function () {
```
lukechampine: our primary style is no space after `function`

PR #220, opened 2016-05-17
Start-up and loading improvements

avahowell:
This pull request improves the functionality of Sia-UI's loading screen and Siad startup behavior.  The core functionality of the loading screen has been changed from being split across electron's `main` and `renderer` processes and now solely exists inside of the `renderer` process.  The IPC polling code used previously has been replaced with async callbacks and promises.  `detached` will also now be set on launch if `siad` is running before the UI is launched.

This should fix #216 as well as the bug mentioned [here](https://bitcointalk.org/index.php?topic=1060294.msg14768506#msg14768506).

PR #221, opened 2016-05-17
Separate siad path and data directory

avahowell:
See https://github.com/NebulousLabs/Nodejs-Sia/pull/26.  This pull depends on #220

### DIFF
--- a/js/mainjs/config.js
+++ b/js/mainjs/config.js
@@ -6,7 +6,8 @@ const Path = require('path');
 const defaultConfig = {
 	homePlugin:  'Overview',
 	siad: {
-    	path: Path.join(__dirname, '../../', 'Sia'),
+    path: Path.join(__dirname, '../../Sia/siad'),
+    datadir: Path.join(__dirname, '../../Sia'),
 		detached: false,
 	},
 	closeToTray: process.platform === 'win32' || process.platform === 'darwin' ? true : false,
--- a/js/mainjs/config.js
+++ b/js/mainjs/config.js
@@ -6,8 +6,8 @@ const Path = require('path');
 const defaultConfig = {
 	homePlugin:  'Overview',
 	siad: {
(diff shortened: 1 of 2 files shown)


js/mainjs/config.js
```diff
@@ -6,7 +6,8 @@ const Path = require('path');
 const defaultConfig = {
 	homePlugin:  'Overview',
 	siad: {
-    	path: Path.join(__dirname, '../../', 'Sia'),
+    path: Path.join(__dirname, '../..', 'Sia', 'siad'),
+    datadir: Path.join(__dirname, '../..', 'Sia'),
```
lukechampine: indentation?
(should have a lint rule for this)

avahowell: There's this in .jshintrc:

```
"indent"        : 4         // Specify indentation spacing
```

Kind of odd that it didn't pick that up.

js/mainjs/config.js
```diff
@@ -6,7 +6,8 @@ const Path = require('path');
 const defaultConfig = {
 	homePlugin:  'Overview',
 	siad: {
-    	path: Path.join(__dirname, '../../', 'Sia'),
+		path: Path.join(__dirname, '../..', 'Sia', 'siad'),
+		datadir: Path.join(__dirname, '../..', 'Sia'),
```
DavidVorick, line 10: These should use multiple params instead of `/`

js/rendererjs/loadingScreen.js
```diff
@@ -84,7 +91,7 @@ module.exports = function(initUI) {
 			defaultPath: Path.join('../', config.path),
```
DavidVorick, line 91: Same here

js/mainjs/config.js
lukechampine, line 10: also, I think `Path.resolve` is preferred over `Path.join` here, though the difference is subtle

PR #223, opened 2016-05-18
Don't use slashes in path.join

avahowell:
Addresses the issues brought up in #221


lukechampine:
what do you think about using `Path.resolve` instead of `Path.join`?

avahowell:
I didn't use it because I found it a bit more ambiguous than explicitly constructing the path using `join` and `__dirname`.

PR #225, opened 2016-05-25
New plugin API specification proposal

avahowell:
This is a proposed specification for a new plugin API for Sia-UI.  I have included the most simple, basic features which I think should be included.  We should discuss the outlined functionality and amend the spec with any additional functionality which we desire.  All built-in plugins will be rewritten to utilize the new plugin API.

### DIFF
--- a/doc/spec/PluginAPI.md
+++ b/doc/spec/PluginAPI.md
@@ -4,15 +4,15 @@
 
 This specification outlines the functionality that Sia-UI's plugin API exposes to developers.
 
-## Desired Functionality
+## Functionality
 
-The plugin API should expose a simple interface for making API calls to siad, creating file dialogs, displaying notifications, and displaying error messages.  This functionality will be wrapped in an object, `SiaAPI`, which is set as a global inside every plugin.  This object will expose an interface to the running, configured, instance of `sia.js` to plugin authors.  It will wrap file dialogs and popup messages with the appropriate IPC calls, so plugin authors do not need to communicate to the main UI directly over ipc.
-
-I envision `SiaAPI` having the following available functions:
+The Sia-UI Plugin api exposes a simple interface for making API calls to siad, creating file dialogs, displaying notifications, and displaying error messages.  This interface is assigned to the `window` object of each plugin, and has the following functions:
 
 - `SiaAPI.call()`, a wrapper to the configured `sia.js`'s `.apiCall` function.
+- `SiaAPI.config`, the current Sia-UI config.
+- `SiaAPI.hastingsToSiacoins`, conversion function from hastings to siacoins.  Returns a `BigNumber` and takes either a `BigNumber` or `string`.
+- `SiaAPI.siacoinsToHastings`, conversion function from siacoins to hastings.
 - `SiaAPI.openFile(options)`, a wrapper which calls Electron.dialog.showOpenDialog with `options`.
 - `SiaAPI.saveFile(options)`, a wrapper which calls Electron.dialog.showSaveDialog with `options`.
-- `SiaAPI.message(options)`, a wrapper which calls Electron.showMessageBox with `options`.
-- `SiaAPI.error(options)`, a wrapper which calls Electron.showErrorBox with `options`.
-
+- `SiaAPI.showMessage(options)`, a wrapper which calls Electron.showMessageBox with `options`.
+- `SiaAPI.showerror(options)`, a wrapper which calls Electron.showErrorBox with `options`.
--- a/doc/spec/PluginAPI.md
+++ b/doc/spec/PluginAPI.md
@@ -6,10 +6,21 @@ This specification outlines the functionality that Sia-UI's plugin API exposes t
 
 ## Desired Functionality
 
-The plugin API should provide an importable javascript module that provides all of the following features:
+The plugin API should be simple to use.  It should expose functionality which has been explicitly defined by the main UI.  This functionality should encompass most of what can be done with the Sia API.  State should be contained within the main UI and flow unidirectionally to the plugins.  Plugins will be assigned a `SiaAPI` global when they are created by the main UI.  This `SiaAPI` global includes functions which dispatch actions, such as
 
-- A function which registers a plugin with Sia-UI.
-- Access to the running, configured, `sia.js` wrapper.
-- Functions to open and save files on the main UI process.
+* `SiaAPI.getWalletStatus()`
+* `SiaAPI.getTransactions()`
 
-The plugin API should be framework and architecture agnostic.  It should not lock developers into any particular code structure, style, or framework.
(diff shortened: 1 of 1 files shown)


doc/spec/PluginAPI.md
```diff
@@ -0,0 +1,16 @@
...
+
+This specification outlines the functionality that Sia-UI's plugin API exposes to developers.
+
+## Desired Functionality
+
+The plugin API should provide an importable javascript module that provides all of the following features:
+
+- A function which registers a plugin with Sia-UI.
+- Access to the running, configured, `sia.js` wrapper.
+- Functions to open and save files on the host.
```
lukechampine: should clarify this language, since "host" in this context does not mean a Sia host

doc/spec/PluginAPI.md
```diff
@@ -0,0 +1,16 @@
...
+- Functions to control the appearance of the plugin.  This includes the ability to set the icon for the plugin and the plugin's order in the sidebar.
```
lukechampine: What's your opinion on providing a function for setting the icon vs. loading it automatically from the plugin folder? My thoughts:
- Function: more flexible than requiring an icon in a specific place; can be (ab)used to make animated icons or notification icons; might be forgotten by plugin author
- File: harder to forget; reduces API surface; not programmable; increases coupling

So I'm leaning towards making it a function, I guess. It feels more explicit that way (less magic), and can be documented along with the rest of the plugin API.

However, I don't think plugins should have control over their ordering in the sidebar. That seems best left to the user.

DavidVorick: Agreed about the ordering, should be user specified.

I... don't see how you would do it as a function. Feels janky.

avahowell: I think it's best if we just assume a static icon location, similar to how favicons work with browsers.  The ordering will be controlled by the main UI, maybe with the ability to drag and slide plugins around the sidebar, or just with an ordering setting in the future `settings` plugin.  I've amended the spec to remove the appearance functions.

doc/spec/PluginAPI.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Introduction
+
+This specification outlines the functionality that Sia-UI's plugin API exposes to developers.
+
+## Desired Functionality
+
+The plugin API should provide an importable javascript module that provides all of the following features:
+
+- A function which registers a plugin with Sia-UI.
```
lukechampine: what is meant by "register?" What would that look like in pseudocode?

johnathanhowell:
I've now rewritten this specification around the idea of using Redux to encapsulate all UI state in the main UI, and providing a simple eventemitter and action dispatcher object for Plugin authors to use in order to implement their desired functionality.

doc/spec/PluginAPI.md
```diff
@@ -0,0 +1,18 @@
...
+
+The plugin API should expose a simple interface for making API calls to siad, creating file dialogs, displaying notifications, and displaying error messages.  This functionality will be wrapped in an object, `SiaAPI`, which is set as a global inside every plugin.  This object will expose an interface to the running, configured, instance of `sia.js` to plugin authors.  It will wrap file dialogs and popup messages with the appropriate IPC calls, so plugin authors do not need to communicate to the main UI directly over ipc.
+
+I envision `SiaAPI` having the following available functions:
+
+- `SiaAPI.call()`, a wrapper to the configured `sia.js`'s `.apiCall` function.
+- `SiaAPI.openFile(options)`, a wrapper which calls Electron.dialog.showOpenDialog with `options`.
+- `SiaAPI.saveFile(options)`, a wrapper which calls Electron.dialog.showSaveDialog with `options`.
+- `SiaAPI.message(options)`, a wrapper which calls Electron.showMessageBox with `options`.
+- `SiaAPI.error(options)`, a wrapper which calls Electron.showErrorBox with `options`.
```
lukechampine: Do `message` and `error` correspond to our notification/error popups? If so, is it necessary to have two different calls, or could we unify them by adding `isError` (or similar) to `options`?

avahowell: `message` and `error` correspond to `Electron`'s `dialog` module: https://github.com/electron/electron/blob/master/docs/api/dialog.md.

From their docs,

```
type String - Can be "none", "info", "error", "question" or "warning". On Windows, "question" displays the same icon as "info", unless you set an icon using the "icon" option.
```

lukechampine: gotcha. Do we currently make use of that functionality anywhere?

avahowell: Not that I'm aware of.

lukechampine: do we plan to? Is there reason to use those dialogs instead of our internal notification/error system?
If not, we should leave it out of the spec, at least for now. Following the v1.0 API guidelines, we should finalize the smallest reasonable set of primitives.

avahowell: Should `message` and `error` be replaced with calls to the notification system?

lukechampine: Hmm. That's a tricky question. On the one hand, standardizing notifications/errors seems like a good thing. On the other hand, there's nothing inherently special about notifications/errors that wouldn't also apply to things like popups. So how much do we provide, and how much do we leave to the plugin author?

My gut says we shouldn't provide anything for now. Later, it might make sense to provide a notification/error system together with popup templates and the like. Perhaps this would even be separate from the plugin API.

How does the new wallet plugin handle notifications/errors?

lukechampine: Aside from that, the only thing I'm iffy on is the magic implicit `SiaAPI` object. I guess it's not much different from `document`, but I'd prefer that it be explicitly imported. I'm not sure there's an easy way to do this, though.

doc/spec/PluginAPI.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Introduction
+
+This specification outlines the functionality that Sia-UI's plugin API exposes to developers.
+
+## Desired Functionality
+
+The plugin API should expose a simple interface for making API calls to siad, creating file dialogs, displaying notifications, and displaying error messages.  This functionality will be wrapped in an object, `SiaAPI`, which is set as a global inside every plugin.  This object will expose an interface to the running, configured, instance of `sia.js` to plugin authors.  It will wrap file dialogs and popup messages with the appropriate IPC calls, so plugin authors do not need to communicate to the main UI directly over ipc.
+
+I envision `SiaAPI` having the following available functions:
```
lukechampine: formalize this a little (no "I") and it'll be good to merge

avahowell:
revised and rebased.

lukechampine:
Later, let's go through the `doc` folder and update the tutorials there.

PR #226, opened 2016-05-26
New wallet plugin

avahowell:
This is a massive PR.  I deleted the old wallet plugin and started from scratch.  [Electron compile](https://github.com/electron/electron-compile) is now being used, giving us have access to ES6 with few changes to the build system.

The new wallet plugin is written using [redux](https://github.com/reactjs/redux) and React.  All state is immutable.  The main application logic lives in `sagas/` and `reducers/`.  `sagas` consume redux actions and perform asynchronous tasks, handling side effects in an elegant way.  `sagas` dispatch `actions`, which are used to update application state in the `reducers/`.  All components are pure functions which return a react component.  Components are hooked up to the state model using by constructing `containers` using `redux-react`.  

I think this structure provides a flexible, reliable, and testable way of building plugins.  This PR only contains the rewritten plugin, which interfaces with the main UI over IPC in the same way the old plugin did.  There will be another PR which updates this plugin to use the new API, along with a PR which converts `Sia-UI` to `eslint`, from `jshint`.  Until then, linting is disabled for this plugin.


lukechampine:
Can you provide screenshots of the new plugin?

avahowell:
screen shot 2016-05-26 at 11 37 42 pm
screen shot 2016-05-26 at 11 37 48 pm
screen shot 2016-05-26 at 11 41 22 pm
screen shot 2016-05-26 at 11 38 25 pm
screen shot 2016-05-26 at 11 38 40 pm
screen shot 2016-05-26 at 11 41 27 pm
screen shot 2016-05-26 at 11 41 31 pm

plugins/Wallet/js/components/lockscreen.js
```diff
@@ -0,0 +1,48 @@
...
+			</div>
+		);
+	}
+	if (unlocking) {
+		return (
+			<div className="modal">
+				<div className="lockscreen">
+					<span> Unlocking your wallet... </span>
+				</div>
+			</div>
```
lukechampine: this doesn't seem very DRY. Maybe:

``` js
if (unlocked && encrypted && !unlocking) {
    return (
        <div></div>
    );
}
var inner;
if (!unlocked && encrypted && !unlocking) {
    inner = <foo>;
} else if (!encrypted && !unlocking) {
    inner = <bar>;
} else if (unlocking) {
    inner = <baz>;
}
return (
    <div className="modal">
        <div className="lockscreen">
            { inner }
        </div>
    </div>
);
```

I guess it depends on whether we expect these to diverge more in the future.

plugins/Wallet/js/components/newwalletdialog.js
```diff
@@ -0,0 +1,31 @@
...
+	if (!visible) {
+		return (
+			<div></div>
+		);
+	}
+	return (
+		<div className="newwallet-dialog">
+			<p> You have created a new wallet!  Please write down the seed and password in a safe place.  If you forget your password, you won't be able to access your wallet. </p>
+			<h2> Seed: </h2>
+			<span className="newwallet-seed">
```
lukechampine: lol. I wonder if it would render properly if we used the `.jsx` extension.

plugins/Wallet/js/components/newwalletdialog.js
```diff
@@ -0,0 +1,31 @@
+import React, { PropTypes } from 'react';
+
+const NewWalletDialog = ({visible, password, seed, actions}) => {
+	if (!visible) {
+		return (
+			<div></div>
```
lukechampine: not sure I like this pattern. Seems like the parent should be the one deciding whether or not to render the child, rather than always rendering a sometimes-invisible component.

avahowell: I agree.  Or at least, it should be consistent.  I will change the pattern to consistently be the parent deciding which elements to render, as is done in `lockscreen.js`

plugins/Wallet/js/index.js
```diff
@@ -0,0 +1,46 @@
...
+	rootReducer,
+	applyMiddleware(sagaMiddleware)
+);
+sagaMiddleware.run(sagas.watchGetLockStatus);
+sagaMiddleware.run(sagas.watchUnlockWallet);
+sagaMiddleware.run(sagas.watchCreateNewWallet);
+sagaMiddleware.run(sagas.watchGetBalance);
+sagaMiddleware.run(sagas.watchGetTransactions);
+sagaMiddleware.run(sagas.watchGetNewReceiveAddress);
+sagaMiddleware.run(sagas.watchSendSiacoin);
```
lukechampine, line 25: seems like the preferred way to handle this is a "root saga," as exemplified here:
https://github.com/yelouafi/redux-saga/blob/master/examples/real-world/index.js
https://github.com/yelouafi/redux-saga/blob/master/examples/real-world/sagas/index.js

plugins/Wallet/js/reducers/newwalletdialog.js
```diff
@@ -0,0 +1,22 @@
...
+	password: '',
+	seed: '',
+});
+
+export default function newwalletdialog(state = initialState, action) {
+	switch (action.type) {
+	case SHOW_NEW_WALLET_DIALOG:
+		return state.set('visible', true)
+								.set('password', action.password)
+								.set('seed', action.seed)
```
lukechampine: some funky formatting. Use tabs for indentation, spaces for alignment! :P

plugins/Wallet/js/reducers/wallet.js
```diff
@@ -0,0 +1,39 @@
...
+	case constants.SET_UNLOCKED:
+		return state.set('unlocked', true)
+								.set('unlocking', false);
+	case constants.SET_ENCRYPTED:
+		return state.set('encrypted', true);
+	case constants.SET_UNENCRYPTED:
+		return state.set('encrypted', false);
+	case constants.SET_BALANCE:
+		return state.set('confirmedbalance', action.confirmed)
+								.set('unconfirmedbalance', action.unconfirmed);
```
lukechampine: formatting

plugins/Wallet/js/sagas/helpers.js
```diff
@@ -0,0 +1,92 @@
...
+	})
+})
+
+// Compute the net value and currency type of a transaction.
+const computeSum = (txn) => {
+	var value = new BigNumber(0);
+	var currency;
+	if (txn.inputs) {
+		for (let i = 0; i < txn.inputs.length; i++) {
+			const input = txn.inputs[i];
```
lukechampine: doesn't js have some sort of range loop?
seems like a `reduce` would be warranted here anyway though

plugins/Wallet/js/sagas/helpers.js
```diff
@@ -0,0 +1,92 @@
...
+			if (output.walletaddress) {
+				value = value.add(output.value);
+			}
+			// TODO: Imperfect way to go about determining currency of
+			// transaction from 'miner', 'siacoin', or 'siafund'
+			currency = output.fundtype.split(' ')[0];
+		}
+	}
+	return {
+		value: Siad.hastingsToSiacoins(value),
```
lukechampine: does this handle negative numbers properly?

plugins/Wallet/js/sagas/helpers.js
```diff
@@ -0,0 +1,92 @@
...
+export const parseRawTransactions = (response) => {
+	var parsedTransactions = List();
+	if (!response.unconfirmedtransactions) {
+		response.unconfirmedtransactions = [];
+	}
+	if (!response.confirmedtransactions) {
+		response.confirmedtransactions = [];
+	}
+	const rawTransactions = response.unconfirmedtransactions.concat(response.confirmedtransactions);
+	for (let i = 0; i < rawTransactions.length; i++ ) {
```
lukechampine: seems like a `map` would be appropriate here

plugins/Wallet/js/sagas/helpers.js
```diff
@@ -0,0 +1,92 @@
...
+		const { value, currency } = computeSum(rawTransactions[i]);
+		let confirmed = true;
+		if (rawTransactions[i].confirmationtimestamp === uint64max) {
+			confirmed = false;
```
lukechampine: prefer `let confirmed = (rawTransactions[i].confirmationtimestamp !== uint64max)`

plugins/Wallet/js/sagas/helpers.js
```diff
@@ -0,0 +1,92 @@
...
+			currency,
+			value: value.round(4).toString(),
+			transactionid: rawTransactions[i].transactionid,
+			confirmationtimestamp: rawTransactions[i].confirmationtimestamp,
+		})
+	}
+	// Return the transactions, sorted by timestamp.
+	// See https://facebook.github.io/immutable-js/docs/#/Iterable/sort for more on how this works.
+	return parsedTransactions.sort((t1, t2) => {
+		if (t1.confirmationtimestamp > t2.confirmationtimestamp) {
```
lukechampine: not sure, but can you use `parsedTransactions.sortBy(txn => txn.confirmationtimestamp)`?

lukechampine:
Spitballing:
After seeing the "send coins" screenshot, it occurred to me that the plugin API should probably allow for fine-grained Siad call error handling. As an example, if you try to send coins and you get an error saying "invalid address," it'd be nice to highlight the address box red.
(I'm not really sure if this falls to the plugin API, or the way calls are handled by sagas inside the plugin)

plugins/Wallet/js/sagas/helpers.js
avahowell: I believe so.  This function was largely lifted out of the old wallet plugin, TODOs included.  `currency` isn't currently used in the UI anyway, however.

lukechampine:
overall this looks quite good. I'm optimistic for the future of the UI :+1:
We could certainly use some documentation on how the react/redux/saga architecture works, for future contributors. But let's delay that until the plugin is done and another plugin or two has been written.

plugins/Wallet/js/reducers/newwalletdialog.js
```diff
@@ -0,0 +1,20 @@
...
+});
+
+export default function newwalletdialog(state = initialState, action) {
+	switch (action.type) {
+	case SHOW_NEW_WALLET_DIALOG:
+		return state.set('password', action.password)
+                .set('seed', action.seed)
+								
+	case DISMISS_NEW_WALLET_DIALOG:
+		return state.set('visible', false);
```
lukechampine: looks like this was left over. Double-check for any other overlooked `visible` vars.

lukechampine:
I am a little concerned about consistency. Sometimes we have trailing commas, sometimes not. Sometimes JSX content has extra spaces around it. Sometimes functions are directly `export`ed, sometimes they are defined and exported separately. Hopefully a proper `.eslintrc` will address most of these irregularities.

avahowell:
Agreed, an `eslintrc` would make a lot of this better.  I added an issue to that effect a few minutes ago.  A proper `eslintrc` will require changes to the overall linting system and code style which I think is out of scope for this PR.

plugins/Wallet/js/containers/newwalletdialog.js
```diff
@@ -0,0 +1,16 @@
+import NewWalletDialogView from '../components/newwalletdialog.js';
+import { dismissNewWalletDialog } from '../actions/wallet.js';
+import { bindActionCreators } from 'redux';
+import { connect } from 'react-redux';
+
+const mapStateToProps = (state) => ({
+	visible: state.newwalletdialog.get('visible'),
```
lukechampine: left over?

plugins/Wallet/js/containers/receiveprompt.js
```diff
@@ -0,0 +1,15 @@
+import ReceivePromptView from '../components/receiveprompt.js';
+import { connect } from 'react-redux';
+import { hideReceivePrompt } from '../actions/wallet.js';
+import { bindActionCreators } from 'redux';
+
+const mapStateToProps = (state) => ({
+	visible: state.receiveprompt.get('visible'),
```
lukechampine: ditto

avahowell: These were removed in 33bec20.

lukechampine: oh? Then why are they still in the diff, eh?

avahowell: I /thought/ these were removed by 33bec20.  I actually removed them in 7558819.

Issue #227, opened 2016-05-27
Code style and linting system should be upgraded


avahowell:
Since we are switching to `electron-compile`, the primary code style should change to the latest ES6 standards, using `eslint` for linting.

PR #229, opened 2016-05-28
Add eslint linting

avahowell:
This PR implements #227.  Right now, everything is ignored except the new wallet plugin.  As remaining code is refactored it will be whitelisted in .eslintignore.


package.json
```diff
@@ -6,11 +6,14 @@
...
     "chai": "^3.4.1",
     "chai-as-promised": "^5.1.0",
     "electron-mocha": "^0.7.0",
     "electron-packager": "^7.0.2",
     "electron-prebuilt-compile": "^1.2.0",
+    "eslint": "^2.11.0",
+    "eslint-plugin-react": "^5.1.1",
     "ink-docstrap": "^1.0.3",
     "jsdoc": "^3.4.0",
     "jshint": "^2.9.1-rc1"
```
lukechampine: can we get ride of some of these dependencies now? Specifically `chai`, `chai-as-promised`, `electron-mocha`, `ink-docstrap`, `jsdoc`, and `jshint`?

avahowell: yes.  I'm going to leave `chai` since I plan to use it to test the Wallet plugin (and the plugin API/new main UI stuff), but the rest can go.

plugins/Wallet/js/components/receiveprompt.js
```diff
@@ -1,15 +1,18 @@
 import React, { PropTypes } from 'react'
 
-const ReceivePrompt = ({address, actions}) => (
-	<div className="modal">
-		<div className="receive-prompt">
-			You can receive Siacoins using the following address:
-			<div className="wallet-address">{ address }</div>
-			<button className="receiveprompt-dismissbtn" onClick={actions.hideReceivePrompt}>OK</button>
+const ReceivePrompt = ({address, actions}) => {
+	const handleDismissClick = () => actions.hideReceivePrompt()
```
lukechampine, line 4: why was this changed to a standalone function? Are we standardizing on "`onClick` handlers should be defined in the Component body?"

avahowell, line 4: Yes.  Right now we're not doing anything except dispatching the action, but in the future we could do more (validation for example).  This is also enforced by `"react/jsx-handler-names": 1` in eslintrc.  The alternative would be to rename all click handler actions with the `handle` prefix, or decide on a different prefix to be standard.

Issue #232, opened 2016-05-31
Upgrade Sia-UI to ES6


avahowell:
Since we've switched to `electron-compile` and `eslint`, the rest of Sia-UI should be upgraded to ES6.

PR #236, opened 2016-06-01
implement eslint and upgrade code style to ES6 standard

avahowell:
Addresses #232


.eslintignore
```diff
@@ -1,2 +1,4 @@
 **/*.js
 !plugins/Wallet/*
+!index.js
+!js/*
```
lukechampine, line 4: so what's actually excluded at this point? Just the non-wallet plugins?

js/mainjs/addIPCListeners.js
```diff
@@ -1,56 +1,53 @@
-'use strict';
-
 // Electron main process libraries
-const Electron = require('electron');
-const Dialog = Electron.dialog;
-const IPCMain = Electron.ipcMain;
-const Menu = Electron.Menu;
+import { dialog, ipcMain, Menu } from 'electron'
```
lukechampine, line 2: what's the rule for capital vs lowercase?

.eslintignore
avahowell, line 4: Yes.

js/mainjs/addIPCListeners.js
avahowell, line 2: ES6 member imports (`import { blah ...}`) import exported members directly by name.  As such, they must match the names defined in the module.

.eslintignore
lukechampine, line 4: there's probably an easier way to specify that...but I'm assuming we won't need an `.eslintignore` soon anyway

PR #239, opened 2016-06-03
New plugin API implementation

avahowell:
This PR implements the API described in #225 .  The pluginFactory system has been pulled out and replaced with with the `plugins.js` module.  The `Host` and `Files` plugins remain broken.  The `Wallet`, `About`, and `Overview` plugins now use the new `SiaAPI`, and most IPC calls have been removed.


js/rendererjs/plugins.js
```diff
@@ -0,0 +1,102 @@
...
+	const unsanitizedFolders = fs.readdirSync(path)
+	for (const p in unsanitizedFolders) {
+		const pluginPath = Path.join(path, unsanitizedFolders[p])
+		try {
+			fs.statSync(Path.join(pluginPath, '/index.html'))
+		} catch (e) {
+			console.error(e)
+			continue
+		}
+		pluginFolders = pluginFolders.push(pluginPath)
```
lukechampine: seems like a good place for a `map` or `filter`: construct the initial list containing all paths, then filter out the ones for which `statSync` fails

js/rendererjs/uiManager.js
```diff
@@ -125,20 +127,41 @@ function checkUpdate() {
...
-	// Wait for the plugin system to load, then call callback
+	let plugins = scanFolder(defaultPluginDirectory)
+
+	// The home plugin should be first in the sidebar.
+	// We probably want a priority system for this instead.
+	plugins = plugins.sort((p1, p2) => {
+		if (getPluginName(p2) === defaultHomePlugin) {
+			return 1
+		}
+		// Sort the rest alphabetically
```
lukechampine, line 144: Ok for now, but I don't think alphabetical ordering is desired. For example, we always want `About` to be last.
How did the old code handle this?

PR #246, opened 2016-06-13
Terminal Plugin

droghio:
A plugin wrapper for `siac` to bring command line support to the UI . Features include command history, integrated help menu (type 'help' or '?' to toggle it), and password prompts. This plugin is made in a similar style to the new wallet plugin #226. All code is written using the `redux-react` framework. A `siac` binary is required, and on first launch the plugin will prompt the user to provide the path to the binary and store this location in the config file for future use. Plugin screenshots are below.

screen shot 2016-06-13 at 7 22 02 pm
screen shot 2016-06-13 at 7 22 12 pm
screen shot 2016-06-13 at 7 22 18 pm


plugins/Terminal/js/reducers/commandline.js
```diff
@@ -0,0 +1,111 @@
...
+	switch (action.type) {
+
+	case constants.ADD_COMMAND:
+			//Add command to the command history.
+		console.log('NEW COMMAND!')
+		var newCommandHistory = state.get('commandHistory').push(action.command)
+		var newState = state.set('commandHistory', newCommandHistory)
+		newState = newState.set('commandIndex', 0)
+		newState = newState.set('currentCommand', '')
+		return newState
```
avahowell: prefer minimal variable declarations in reducers, ie

```
return state.set('commandHistory', state.get('commandHistory').push(action.command))
                .set('commandIndex', 0)
                .set('currentCommand', '')
```

plugins/Terminal/js/reducers/commandline.js
```diff
@@ -0,0 +1,111 @@
...
+
+
+	case constants.UPDATE_COMMAND:
+			//Updates output of command given by command name and id.
+		var newCommandHistory = state.get('commandHistory')
```
avahowell: newCommandHistory is redeclared from the previous case here

plugins/Terminal/js/components/seedprompt.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react'
+const querystring = require('querystring')
+
+const WalletSeedPrompt = ({ showSeedPrompt, currentCommand, actions }) => {
+	componentDidUpdate: {
```
avahowell: I'm not sure I understand the need for componentDidUpdate here.

plugins/Terminal/js/components/walletpasswordprompt.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react'
+const querystring = require('querystring')
+
+//This command needs a second prompt.
+const moreSpecialCommands= [ ['wallet', 'load', 'seed'] ]
+
+const WalletPasswordPrompt = ({ showWalletPrompt, currentCommand, actions }) => {
+	componentDidUpdate: {
```
avahowell: ditto

plugins/Terminal/js/index.js
```diff
@@ -0,0 +1,224 @@
+// index.js: main entrypoint for the Sia-UI wallet plugin.
+import React from 'react'
+import ReactDOM from 'react-dom'
+import { createStore } from 'redux'
+import { Provider } from 'react-redux'
+import rootReducer from './js/reducers/index.js'
+import CommandLine from './js/components/app.js'
+import Path from 'path'
+const fs = require('fs')
```
avahowell: should use `import`

plugins/Terminal/js/index.js
```diff
@@ -0,0 +1,224 @@
...
+		defaultPath: Path.join('..', SiaAPI.config.attr('siac').path || './' ),
+		filters: [{ name: 'siac', extensions: ['*'] }],
+	})
+	SiaAPI.config.attr('siac', { path: Path.dirname(siacPath[0]) })
+	SiaAPI.config.save()
+})
+
+const isCommandSpecial = function(commandString, specialArray) {
+	//Cleans string and sees if any subarray in array starts with the string when split.
+	var args = commandString.replace(/\s*\s/g, ' ').trim().split(' ')
```
avahowell: `args` is never reassigned, prefer const

plugins/Terminal/js/index.js
```diff
@@ -0,0 +1,224 @@
...
+		title: 'Please locate siac.',
+		properties: ['openFile'],
+		defaultPath: Path.join('..', SiaAPI.config.attr('siac').path || './' ),
+		filters: [{ name: 'siac', extensions: ['*'] }],
+	})
+	SiaAPI.config.attr('siac', { path: Path.dirname(siacPath[0]) })
+	SiaAPI.config.save()
+})
+
+const isCommandSpecial = function(commandString, specialArray) {
```
avahowell: isCommandSpecial is never used in index.js, I think it's accessible from the global scope though.  It should be in its own module and explicitly exported.

plugins/Terminal/js/index.js
```diff
@@ -0,0 +1,224 @@
...
+		hostname: siaAddr.hostname,
+		port: siaAddr.port,
+		path: apiURL,
+		method: 'POST',
+		headers: {
+			'User-Agent': 'Sia-Agent',
+			'Content-Type': 'application/x-www-form-urlencoded',
+		},
+	}
+	var req = http.request(options, (res) => {
```
avahowell: use `SiaAPI.call` to make requests to Siad, it exposes the active, configured, instance of siad that the main UI is using.

plugins/Terminal/js/reducers/commandline.js
```diff
@@ -0,0 +1,111 @@
...
+			newCommand = newCommand.set('stat', action.stat)
+		}
+
+		console.log(newCommand)
+		newCommandHistory = newCommandHistory.set(commandIdx, newCommand)
+		return state.set('commandHistory', newCommandHistory)
+
+
+	case constants.LOAD_PREV_COMMAND:
+		var newCommandIndex = state.get('commandIndex')
```
avahowell: newCommandIndex is redeclared also

avahowell:
needs updated `.eslintignore` to enable eslint on this plugin

plugins/Terminal/js/reducers/commandline.js
```diff
@@ -0,0 +1,111 @@
...
+			newCommand = newCommand.set('result', newCommand.get('result') + action.dataChunk)
+		}
+
+		if (action.stat) {
+			newCommand = newCommand.set('stat', action.stat)
+		}
+
+		console.log(newCommand)
+		newCommandHistory = newCommandHistory.set(commandIdx, newCommand)
+		return state.set('commandHistory', newCommandHistory)
```
avahowell: I'm not sure I like handling these chains of side effects in the reducers like this.  I used `redux-saga` in the Wallet plugin to implement this kind of logic.  @lukechampine, thoughts?

lukechampine: Which side effects are you referring to, exactly? The console logging definitely doesn't belong, but the rest seems pure.

plugins/Terminal/css/style.css
```diff
@@ -0,0 +1,181 @@
...
+.command-history-list {
+    width: 100%;
+    /* Size of main section minus footer. */
+    height: calc( 100% - 2.4em - 9px );
+    max-height: calc( 100% - 2.4em - 9px );
+    overflow-y: scroll;
+    transition: all .5s;
+}
+.overview .command-history-list {
+    /* When overview is openned make this the size of the screen minus footer minus size of overview pane. */
```
lukechampine: typo

plugins/Terminal/css/style.css
```diff
@@ -0,0 +1,181 @@
...
+    position: relative;
+    width: calc( 100% - 1.2em - 2px );
+    margin: .5em;
+}
+
+
+.modal {
+    font-family: 'Roboto Condensed', sans-serif;
+	position: absolute;
+	top: 0;
```
lukechampine, line 149: the indentation here is all over the place. (In many other files as well.) There are a few other stylistic issues I noticed, like `.command-overview p{` (missing a space after `p`) and 3+ newlines between declarations (generally should be limited to just 1).
@johnathanhowell, is there some gofmt-style tool for js that can automatically correct these things?

plugins/Terminal/js/actions/commandline.js
```diff
@@ -0,0 +1,47 @@
...
+	type: constants.HIDE_WALLET_PROMPT,
+})
+
+
+export const showSeedPrompt = () => ({
+	type: constants.SHOW_SEED_PROMPT,
+})
+export const hideSeedPrompt = () => ({
+	type: constants.HIDE_SEED_PROMPT,
+})
```
lukechampine, line 44: I've waffled on this sort of action before: should it be two separate actions, or just one action that takes a bool? @johnathanhowell, what's your preference?

plugins/Terminal/js/components/commandhistorylist.js
```diff
@@ -0,0 +1,41 @@
...
+	componentDidUpdate: {
+		var ch = document.getElementsByClassName('command-history-list')[0]
+		if (ch) {
+			//Give the DOM time to process the DOM changes.
+			setTimeout( () => ch.scrollTop = ch.scrollHeight, 0)
+		}
+	}
+
+	render: {
+		console.log('Re-rendering.')
```
lukechampine: is there some way to only enable these logging statements during debugging?

plugins/Terminal/css/style.css
avahowell, line 149: not that I'm aware of.  `eslint --fix` will do that for the javascript.

plugins/Terminal/js/components/commandhistorylist.js
```diff
@@ -0,0 +1,41 @@
...
+import { List } from 'immutable'
+
+const CommandHistoryList = ({commandHistory}) => {
+
+	componentDidUpdate: {
+		var ch = document.getElementsByClassName('command-history-list')[0]
+		if (ch) {
+			//Give the DOM time to process the DOM changes.
+			setTimeout( () => ch.scrollTop = ch.scrollHeight, 0)
+		}
```
lukechampine: This is manually editing the DOM, right? Seems very un-idiomatic for a React app. Ideally, all your components should be pure functions of their props.

plugins/Terminal/js/actions/commandline.js
avahowell, line 44: I've been leaning more towards the `toggle` pattern where it makes sense, ie:

```
export const toggleDownloadList = () => ({
    type: constants.TOGGLE_DOWNLOADS_LIST,
})
```

... in the reducer:

```
case constants.TOGGLE_DOWNLOADS_LIST:
        return state.set('showDownloadList', !state.get('showDownloadList'))
```

there are places where having separate `hide` and `show` actions are fine.  I prefer that to a bool.

plugins/Terminal/js/components/commandinput.js
```diff
@@ -0,0 +1,75 @@
+import React from 'react'
+import { Map } from 'immutable'
+
+//These commands need a password prompt.
+const specialCommands = [ ['wallet', 'unlock'], ['wallet', 'load', 'seed'], ['help'], ['?'] ]
```
lukechampine: comment is not totally accurate -- `help` and `?` are just special.

plugins/Terminal/js/components/commandinput.js
```diff
@@ -0,0 +1,75 @@
...
+
+const CommandInput = ({currentCommand, showCommandOverview, actions}) => {
+	componentDidUpdate: {
+		setTimeout(() => document.getElementById('command-input').focus(), 0)
+	}
+
+	reduce: {
```
lukechampine: I'm confused by this syntax. Doesn't this need to be named `render`? And doesn't it need to be a function? This just looks like an object.

plugins/Terminal/js/components/commandinput.js
```diff
@@ -0,0 +1,75 @@
...
+				actions.loadPrevCommand(e.target)
+				setTimeout( () => {
+					var commandinput = document.getElementById('command-input')
+					commandinput.setSelectionRange(commandinput.value.length, commandinput.value.length)
+				}, 0)
+			} else if (e.keyCode === 40) {
+				//Down arrow.
+				actions.loadNextCommand(e.target)
+				setTimeout(() => {
+					var commandinput = document.getElementById('command-input')
```
lukechampine: `document` should almost never appear in a React component. It's a sign that you aren't passing the component enough props, or that you don't have enough actions.

plugins/Terminal/js/components/commandinput.js
```diff
@@ -0,0 +1,75 @@
...
+				case 3: //?
+					var newText = text || '?'
+					if (showCommandOverview) {
+						actions.hideCommandOverview()
+					} else {
+						actions.showCommandOverview()
+					}
+
+					//The command history won't actually show a help command so it is fine to add the command.
+					var newCommand = Map({ command: newText, result: '', id: Math.floor(Math.random()*1000000) })
```
lukechampine: why is the random id necessary? And why add the command at all, if it won't be displayed?

plugins/Terminal/js/components/commandinput.js
```diff
@@ -0,0 +1,75 @@
...
+		}
+
+		const handleKeyboardPress = (e) => {
+
+			//Enter button.
+			if (e.keyCode === 13) {
+
+				//Check if command is special.
+				switch (  isCommandSpecial(currentCommand, specialCommands) ) {
+				case -1: //Regular command.
```
lukechampine: magic return values are bad. Instead of `isCommandSpecial` returning a number, maybe have a `commandType` function that returns a constant.

This may also not be the right place for your command-handling logic. Generally, components are only concerned with presentation. Their `onChange` functions are passed as props, unless they are very simple. @johnathanhowell can probably advise you better on this.

plugins/Terminal/js/reducers/commandline.js
avahowell: I was referring to the changing of multiple state properties based on somewhat complex logic inside a `case`, but after a second read there aren't any actual side effects.

plugins/Terminal/js/components/commandline.js
```diff
@@ -0,0 +1,47 @@
...
+							version     Print version information<br />
+							stop        Stop the Sia daemon<br />
+							host        Perform host actions<br />
+							hostdb      View or modify the host database<br />
+							miner       Perform miner actions<br />
+							wallet      Perform wallet actions<br />
+							renter      Perform renter actions<br />
+							gateway     Perform gateway actions<br />
+							consensus   Print the current state of consensus<br /><br />
+							Use '[command] --help' for more information about a command.<br />
```
lukechampine, line 35: this seems like it ought to be a table. Though I suppose it's not a big deal as long as it's monospaced. Fits the aesthetic of the plugin.

plugins/Terminal/js/components/commandinput.js
avahowell: +1 this component should be passed an `input` prop and an action that sets the value of that input when it changes.

plugins/Terminal/js/index.js
```diff
@@ -0,0 +1,224 @@
...
+		 if (!err) {
+			resolve()
+		 } else {
+			reject(err)
+		 }
+	})
+})
+
+checkSiaPath().catch(() => {
+	// config.path doesn't exist.	Prompt the user for siad's location
```
lukechampine: should be `siac`

plugins/Terminal/js/index.js
```diff
@@ -0,0 +1,224 @@
...
+		commandString = commandString.slice(6).trim()
+	}
+
+	//Add address flag to siac.
+	var args = commandString.split(' ')
+	if (args.indexOf('-a') === -1 && args.indexOf('--address') === -1 && SiaAPI.config.attr('address')) {
+		args = args.concat([ '-a', SiaAPI.config.attr('address') ])
+	}
+
+	var siac = child_process.spawn('./siac', args, { cwd: SiaAPI.config.attr('siac').path })
```
lukechampine: as I understand it, this means that commands will be run from wherever the siac binary is. This may be important for commands like `siac renter download`, which takes a path as an argument. Not sure what the best way to handle this is.

plugins/Terminal/js/reducers/commandline.js
```diff
@@ -0,0 +1,111 @@
...
+		}
+		return returnvalue
+
+
+	case constants.LOAD_NEXT_COMMAND:
+		var newCommandIndex = state.get('commandIndex')
+		newCommandIndex--
+		if (newCommandIndex < 0) {
+			newCommandIndex = 0
+		}
```
lukechampine: use `Math.min` for these

lukechampine:
Commit messages should be in the imperative; `Add` instead of `Adding`, `Fix` instead of `Fixed`, etc.
They should also not end in a period, and should be less than 50 chars. You can add more detail in the body. See http://chris.beams.io/posts/git-commit
(The only rule we don't particularly care about is capitalizing the subject line)

plugins/Terminal/js/components/commandinput.js
avahowell: Components should be concerned with presentation and basic event functionality.  I'm fine with having a `handleKeyPress` event handler inside a component, but the command spawning logic should be somewhere else.  It's an ideal use case for a saga, ie

```
function *runCommandSaga(action) {
    while (true) {
        const command = yield take(constants.RUN_COMMAND)
        const result = yield spawnCommand(command)
        yield put(receiveResult(result))
    }
}
```

avahowell:
Also, prefer `const` to `let`, and never use var.  Always use `let`.  Might add a linting rule for this

plugins/Terminal/js/actions/commandline.js
lukechampine, line 44: hmm. I'm not sure I like `toggle`; it subtly binds your action to the current state of the program. That is, you need to know what state the list is in when you toggle it. Seems safer to explicitly `show` or `hide` it. (For example, what if two concurrent actions both try to toggle an element?)

plugins/Terminal/js/components/commandinput.js
lukechampine: @droghio, I think you will learn a lot about The React Way™ by refactoring this plugin to purge all instances of `document` from your components.

plugins/Terminal/js/components/commandinput.js
droghio: This syntax is closer to react's class structure: https://facebook.github.io/react/docs/component-specs.html#updating-componentdidupdate

The idea is to auto focus the proper text bar after react redraws the dom so the user doesn't have to keep clicking text boxes. You are right about the `render` vs `reduce` function; I'm not entirely sure why `reduce` is working but I'll make the change. @johnathanhowell recommended I use event listeners instead and I'm looking into it.

plugins/Terminal/js/components/commandinput.js
droghio: The ui updating and command dispatching is fully async and the ui currently allows multiple commands to be updating simultaneously. When a command sends output the id is used to identify and append the output to the proper object in the redux state. This prevents issues that could occur if, for example, `version` command was run twice in a row but both returned at the same time. I wanted to make sure we didn't end up with one displaying nothing and the other displaying "version 1.00version1.00" or something of the sort.

I could sequentially number commands and use this as the id but I was worried that may introduce subtle race-time issues. If react/electron don't mess with the atomic nature of js it may not be an issue. An alternative is to lock the ui input when running a command so only one command could receive output at a time. Might be an issue, ie lock up the ui, for long running commands like `wallet unlock` or if a command doesn't quit correctly.

As for adding the command, this is so that `help` and `?` appear in the command history. It isn't displayed but allows the user to hit the up arrow to pull up the command.

plugins/Terminal/js/index.js
droghio: Hmm good point. I assumed they would use absolute paths when feeding arguments, but that may be a bad assumption.

plugins/Terminal/js/components/commandinput.js
droghio: Let the purge begin.

plugins/Terminal/js/components/seedprompt.js
droghio: Idea was to auto focus to the most prominent textbox automatically. For the wallet/seed modals those textboxes needed to be drawn first. Using `componentDidUpdate` allowed me to schedule the focus call to after the modals were visible. You mentioned event listeners might be a better way to handle this.

plugins/Terminal/css/style.css
droghio, line 149: I forgot eslint doesn't lint css and other assets. If there is a preferred tool I'll setup linting otherwise I will normalize the formatting by hand.

plugins/Terminal/js/index.js
droghio: Nice, that would simplify the code. My only concern is that would disallow the use of the `-a`/`--address` flag to be passed to these functions. Is that something we are ok with losing?

lukechampine: Good point, but I think the terminal plugin is not intended to talk to a siad other than the one the UI is talking to.

plugins/Terminal/js/index.js
lukechampine: Users should rarely be using these commands from the terminal, so I'm not too worried about it. It would be a pain anyway, since the terminal plugin doesn't have filename autocomplete like a real terminal does.

plugins/Terminal/js/index.js
droghio: Ok, if that's the case what should be done about those flags (`-a`, `--address`). Siac currently shows them as global flag options and other commands will accept them. Should I strip out the flags and display an error message if someone attempts to use them?

plugins/Terminal/js/components/commandinput.js
DavidVorick: On some planets, it's just called Murder Night.

plugins/Terminal/js/index.js
lukechampine: hmm. If the direct API call workaround wasn't needed, I don't see any problem with allowing `-a`. It's a little weird, but it shouldn't be disallowed. So I think it's fine to not filter them. If anything, I suppose you could detect if they are being used with the special cases and throw a warning. But I don't think it's a big deal.

plugins/Terminal/js/utils/helpers.js
```diff
@@ -0,0 +1,260 @@
...
+		hostname: siaAddr.hostname,
+		port: siaAddr.port,
+		path: apiURL,
+		method: 'POST',
+		headers: {
+			'User-Agent': 'Sia-Agent',
+			'Content-Type': 'application/x-www-form-urlencoded',
+		},
+	}
+	const req = http.request(options, (res) => {
```
droghio, line 191: @lukechampine, @johnathanhowell, just wanted to confirm that I should remove the http request for the SiadAPI.call. The current http call can handle `--address` attributes and will auto link to the running siad instance otherwise. I'm not sure what we gain by using the SiadAPI, but it would be simple drop in.

lukechampine, line 191: yes. `SiadAPI` means you don't need to load the config.

plugins/Terminal/js/components/commandhistorylist.js
```diff
@@ -0,0 +1,32 @@
+import React, { PropTypes } from 'react'
+import { List } from 'immutable'
+
+export default class CommandHistoryList extends React.Component {
+	componentDidUpdate() {
+		this._commandHistoryList.scrollTop = this._commandHistoryList.scrollHeight
+	}
+
+	render() {
+		console.log('Re-rendering.')
```
lukechampine: please strip out the console logging, here and elsewhere

plugins/Terminal/js/constants/helper.js
```diff
@@ -0,0 +1,5 @@
+export const REGULAR_COMMAND = -1
+export const WALLET_SEED = 0
+export const WALLET_UNLOCK = 1
+export const HELP = 2
+export const HELP_QMARK = 3
```
lukechampine, line 5: not really what I had in mind... the values returned by `commandType` are still "magic."

But if you revisit this code in a month, I'm sure you'll be brimming with ways to improve it, so lets not sweat the small stuff for now.

Issue #248, opened 2016-06-18
A JavaScript error occurred in the main process


roymj88:
Hi,

Tried installing using the steps given:

```
git clone https://github.com/NebulousLabs/Sia-UI.git
npm install
npm start
```

But it shows the following error:

```
Uncaught Exception:
TypeError: protocol.interceptBufferProtocol is not a function
    at initializeProtocolHook (/home/roy/projects/Sia-UI/node_modules/electron-compile/lib/protocol-hook.js:115:12)
    at EventEmitter.protoify (/home/roy/projects/Sia-UI/node_modules/electron-compile/lib/config-parser.js:317:9)
    at emitOne (events.js:101:20)
    at EventEmitter.emit (events.js:188:7)
```

Note: I am on ubuntu with node version `v6.2.2` and npm version`3.9.5`

Kindly advice as to what should be done.

![sc](https://cloud.githubusercontent.com/assets/2633799/16172663/6dc590d8-35aa-11e6-8139-3e5d15d519bf.png)

SonicWizard:
I get the same thing.

avahowell:
I'm pretty sure this is an issue in the upstream package `electron-prebuilt-compile`, as such I've made an issue here: https://github.com/electron/electron-compile/issues/107.

As of right now we've moved the build system to webpack.  This issue should be resolved, so I'm closing it.

roymj88:
@johnathanhowell : So i just pull the latest and things should be fine? Please advice.

avahowell:
Yes, pull the latest master and 

`npm install`
`npm run start`

PR #249, opened 2016-06-20
Implement webpack build system

avahowell:
This PR changes the build system to use webpack.  This should fix the issues reported by people in #248, as `electron-compile` is no longer used.  This PR also means that all dependencies are flattened for release in `dist/`, fixing the long running issue with path length on windows caused by distributing the full `node_modules` folder.


plugins/Wallet/index.html
```diff
@@ -12,6 +12,6 @@
 	<body>
 		<!-- React root -->
 		<div id="react-root"></div>
-		<script src="js/index.js"></script>
+		<script src="../../dist/plugins/Wallet.js"></script>
```
lukechampine, line 15: I worry that this will not be very intuitive for plugin developers.

webpack.config.js
```diff
@@ -0,0 +1,54 @@
...
+	},
+	resolve: {
+		root: path.resolve('./node_modules')
+	}, 
+	resolveLoader: {
+		root: path.resolve('./node_modules'),
+	},
+	node: {
+ 		__dirname: false,
+  	__filename: false,
```
lukechampine: indentation

plugins/Wallet/index.html
avahowell, line 15: I agree, although they do not have to use the `webpack` build system.

lukechampine, line 15: could we modify it such that the compiled js was placed inside the plugin directory, instead of a globally shared directory?

PR #250, opened 2016-06-20
v1.0-rc Files Plugin

avahowell:
This is the new v1.0 renter plugin.

Screenshots:

screen shot 2016-06-20 at 2 43 11 pm
screen shot 2016-06-20 at 12 43 05 pm
screen shot 2016-06-20 at 12 45 10 pm
screen shot 2016-06-20 at 12 45 32 pm
screen shot 2016-06-20 at 12 45 01 pm
screen shot 2016-06-20 at 12 45 04 pm


lukechampine:
91 commits wew lad
net-negative though :+1:

plugins/Files/js/components/app.js
```diff
@@ -0,0 +1,31 @@
+import React, { PropTypes } from 'react'
+import FileBrowser from '../containers/filebrowser.js'
+import AllowanceDialog from '../containers/allowancedialog.js'
+
+const FilesApp = ({unlocked, showAllowanceDialog}) => {
+	let fileBrowserContent
+	if (!unlocked) {
+		fileBrowserContent = (
```
lukechampine: not sure how I feel about this pattern. I think I would prefer to stick everything in the final JSX element rather than assigning to a local variable. If you applied that here, the component would be a single expression.

plugins/Files/js/components/deletedialog.js
```diff
@@ -0,0 +1,29 @@
...
+const DeleteDialog = ({siapath, actions}) => {
+	const onYesClick = () => {
+		actions.deleteFile(siapath)
+		actions.hideDeleteDialog()
+	}
+	const onNoClick = () => actions.hideDeleteDialog()
+	return (
+		<div className="modal">
+			<div className="delete-dialog">
+				<h3> Confrim Deletion </h3>
```
lukechampine: typo

lukechampine:
off-hand, I like how well this plugin showcases the redux architecture. You don't even have to look at the file contents: it's plain to see that there are a bunch of components, slightly fewer containers to combine and compose the components, even fewer reducers to manage the stateful containers, and finally just a handful of sagas for the async reducers.

plugins/Files/js/components/file.js
```diff
@@ -0,0 +1,30 @@
+import React, { PropTypes } from 'react'
+
+const File = ({filename, filesize, onDownloadClick, onDeleteClick}) => (
+	<li>
+		<div className="filename">
+			<i className="fa fa-file"></i>
+			<div className="name">{filename}</div>
+		</div>
+		<div className="file-info">
+			<span className="filesize">{Math.floor(filesize/1000000)} MB</span>
```
lukechampine: we'll want something more sophisticated here, analogous to [how siac does it](https://github.com/NebulousLabs/Sia/blob/master/siac/parse.go#L16). MB aren't suitable for everything, and it looks like this code will report 0MB if the file is less than a megabyte.

plugins/Files/js/components/filebrowser.js
```diff
@@ -0,0 +1,59 @@
...
+	return (
+		<div className="file-browser-container">
+			<div className="file-browser" onDragOver={onDragOver} onMouseLeave={onDragLeave} onDrop={onDrop}>
+				{showUploadDialog ? <UploadDialog /> : null}
+				{showDeleteDialog ? <DeleteDialog /> : null}
+				{dragging ? <DragOverlay /> : null}
+
+				<div className="files-toolbar">
+					<div className="files-usage-info">
+						Storage Usage: {diskusage/1000000000} GB
```
lukechampine: ditto

plugins/Files/js/components/filelist.js
```diff
@@ -0,0 +1,68 @@
+import React, { PropTypes } from 'react'
+import { List } from 'immutable'
+import File from './file.js'
+import Directory from './directory.js'
+import Path from 'path'
+import SearchField from '../containers/searchfield.js'
+
+const FileList = ({files, searchResults, path, showSearchField, actions}) => {
+
+	const onDirectoryClick = (directory) => () => actions.setPath(path + directory.name + '/')
```
lukechampine: shouldn't this be a `path.join`? :P

lukechampine: part of me thinks you should combine this and `onBackClick` in a new action, `actions.cd(dir)`, where the path manipulation logic would live.

plugins/Files/js/components/filelist.js
```diff
@@ -0,0 +1,68 @@
...
+			})
+			actions.downloadFile(file.siapath, Path.join(downloadpath[0], Path.basename(file.siapath)))
+		}
+		const onDeleteClick = () => actions.showDeleteDialog(file.siapath)
+		return (
+			<File key={key} filename={file.name} filesize={file.size} onDownloadClick={onDownloadClick} onDeleteClick={onDeleteClick} />
+		)
+	})
+	console.log(fileElements)
+	console.log(files)
```
lukechampine: left over from debugging?

plugins/Files/js/components/filelist.js
```diff
@@ -0,0 +1,68 @@
...
+		}
+		actions.setPath(newpath)
+	}
+
+	let filelistFiles
+	if (showSearchField) {
+		filelistFiles = searchResults
+	} else {
+		filelistFiles = files
+	}
```
lukechampine, line 27: does it make sense to combine these in a single component? Maybe. Alternatively, search could be its own component that wraps `FileList` (or something like it), providing it with the files returned by the search

plugins/Files/js/components/filelist.js
```diff
@@ -0,0 +1,68 @@
...
+	})
+	console.log(fileElements)
+	console.log(files)
+
+	return (
+		<div className="file-list">
+			{showSearchField ? <SearchField /> : null}
+			<ul>
+				{path !== '' ? <li onClick={onBackClick}><div><i className="fa fa-backward"></i>Back</div></li> : null}
+				{fileElements.size > 0 ? fileElements : <h2> No files have been uploaded! </h2>}
```
lukechampine: won't this display `No files have been uploaded!` if your search returns nothing? Misleading...

plugins/Files/js/components/progressbar.js
```diff
@@ -0,0 +1,21 @@
+import React, { PropTypes } from 'react'
+
+const ProgressBar = ({progress}) => {
+	const style = {
+		width: progress.toString() + '%',
```
lukechampine, line 5: neat!

plugins/Files/js/components/filetransfers.js
```diff
@@ -0,0 +1,41 @@
...
+	let uploadlist
+	if (uploads.size > 0) {
+		uploadlist = (
+			<div className="downloads">
+				<h3> Uploads </h3>
+				<TransferList transfers={uploads} />
+			</div>
+		)
+	}
+	let downloadlist
```
lukechampine: is the default value `null` or `undefined`? What happens if you try to render `undefined`?
either way, might want to follow the ternary pattern used elsewhere.

plugins/Files/js/components/filelist.js
avahowell: Yeah, that is misleading.  I changed the ternary to include `&& !showSearchField` to avoid this happening.

plugins/Files/js/components/filetransfers.js
avahowell: the default value is `undefined`.  Rendering `undefined` has the same effect as rendering `null`, but I agree with using the ternary pattern.  I'm going to separate `uploadlist` and `downloadlist` into their own components and ternary them in `FileTransfers`.

plugins/Files/js/components/app.js
avahowell: agreed, I changed this to be a single expression and also added styling ("You must unlock..." was unstyled previously.)

plugins/Files/js/index.js
```diff
@@ -0,0 +1,38 @@
...
+
+store.dispatch(getWalletLockstate())
+store.dispatch(getFiles())
+store.dispatch(setPath(''))
+store.dispatch(getUploads())
+
+const downloadStart = Date.now()
+
+setInterval(() => {
+	store.dispatch(getDownloads(downloadStart))
```
lukechampine, line 33: The `downloadStart` construction is a little odd. As I understand, it's used to filter out downloads that were not initiated during the current session. Is that the behavior we want, though? In Firefox, I can always view all of the downloads I have ever made.

plugins/Files/js/reducers/files.js
```diff
@@ -0,0 +1,71 @@
...
+	downloading: List(),
+	path: '',
+	searchText: '',
+	uploadSource: '',
+	showAllowanceDialog: false,
+	showUploadDialog: false,
+	showSearchField: false,
+	showFileTransfers: false,
+	showDeleteDialog: false,
+	dragging: false,
```
lukechampine, line 22: wowza that's a lotta state. We can't eliminate it -- this plugin has a lot of inherent state -- but we can break this apart into smaller sub-components. (Later.)

plugins/Files/js/reducers/files.js
```diff
@@ -0,0 +1,71 @@
...
+	console.log(action)
+	switch (action.type) {
+	case constants.RECEIVE_DISK_USAGE:
+		return state.set('usage', action.usage)
+	case constants.RECEIVE_FILES:
+		return state.set('files', action.files)
+                .set('workingDirectoryFiles', ls(action.files, state.get('path')))
+	case constants.SET_SEARCH_TEXT:
+		return state.set('searchResults', searchFiles(state.get('files'), action.text, state.get('path')))
+                .set('searchText', action.text)
```
lukechampine: is there an alignment rule for these? Aligning the `.set`s seems optimal to me.
Also, the nested calls can become hard to read. No shame in splitting it across two lines:

``` js
let results = searchFiles(state.get('files'), action.text, state.get('path'))
return state.set('searchResults', results)
            .set('searchText', action.text)
```

plugins/Files/js/sagas/files.js
```diff
@@ -0,0 +1,227 @@
+import { takeEvery } from 'redux-saga'
+import { put, take, fork, cancel } from 'redux-saga/effects'
+import Path from 'path'
+import * as actions from '../actions/files.js'
+import * as constants from '../constants/files.js'
+import { List } from 'immutable'
+import { sendError, siadCall, totalUsage, readdirRecursive, parseDownloads, parseUploads, estimatedStoragePriceGBSC } from './helpers.js'
+
+const blockMonth = 4382
```
lukechampine, line 9: The figure we use in the backend (and siac) is 4320 blocks per month

plugins/Files/js/sagas/files.js
```diff
@@ -0,0 +1,227 @@
...
+	try {
+		let response = yield siadCall('/renter/contracts')
+		if (!response.contracts) {
+			return
+		}
+		const initialContracts = response.contracts.length
+		while (true) {
+			response = yield siadCall('/renter/contracts')
+			const deltaContracts = response.contracts.length - initialContracts
+			if (deltaContracts === allowanceHosts) {
```
lukechampine: so this will only exit when an additional 24 contracts are formed? (as opposed to 24 total)
hmm. Not sure if that's the correct behavior.

At the very least, in these circumstances you should generally avoid `==` in favor of `>=`

plugins/Files/js/sagas/files.js
```diff
@@ -0,0 +1,227 @@
...
+function *uploadFolderSaga(action) {
+	try {
+		const files = readdirRecursive(action.source)
+		const folderName = Path.basename(action.source)
+		const siafiles = files.map((file) => ({
+			siapath: file.substring(file.indexOf(folderName), file.lastIndexOf(Path.basename(file))),
+			source: file,
+		}))
+		for (let i = 0; i < siafiles.size; i++) {
+			yield put(actions.uploadFile(siafiles.get(i).siapath, siafiles.get(i).source))
```
lukechampine: why isn't this a `foreach` or something?
seems like you could even combine it with the previous loop

plugins/Files/js/sagas/helpers.js
```diff
@@ -0,0 +1,161 @@
...
+		readablesize /= 1000
+	}
+	return readablesize.toFixed(2).toString() + ' ' + readableunit
+}
+
+// return a list of files filtered with path.
+// ... it's ls.
+export const ls = (files, path) => {
+	let fileList = files.filter((file) => file.uploadprogress >= 100)
+	fileList = fileList.filter((file) => file.siapath.indexOf(path) !== -1)
```
lukechampine: combine these

plugins/Files/js/index.js
avahowell, line 33: I went back and forth on this behavior a few times.  It should probably be filterable with a dropdown/checkbox (downloads this session, all downloads)

plugins/Files/js/sagas/helpers.js
```diff
@@ -0,0 +1,161 @@
...
+		let type = 'file'
+		const relativePath = file.siapath.substring(path.length, file.siapath.length)
+		let filename = Path.basename(relativePath)
+		let uploadprogress = Math.floor(file.uploadprogress)
+		if (relativePath.indexOf('/') !== -1) {
+			type = 'directory'
+			filename = relativePath.split('/')[0]
+			uploadprogress = ''
+		}
+		parsedFiles = parsedFiles.set(filename, {
```
lukechampine, line 57: am I wrong in thinking this should be a `.map` instead of a `forEach`? Is it because you might overwrite a filename?

plugins/Files/js/reducers/files.js
avahowell: agreed, sublime shows the .sets as aligned but they need to have all spaces to be aligned properly iirc

plugins/Files/js/sagas/files.js
avahowell: This will exit when either:
- 24 contracts are formed since the saga was started
- the call to `/renter` completes

See https://github.com/NebulousLabs/Sia-UI/pull/250/files#diff-fa3e35a6345aea46e8c098f3dba80505R194 for the cancellation behaviour

plugins/Files/js/sagas/helpers.js
```diff
@@ -0,0 +1,161 @@
...
+		filelist = files
+	}
+	dirfiles.forEach((file) => {
+		const filepath = Path.join(path, file)
+		const stat = fs.statSync(filepath)
+		if (stat.isDirectory()) {
+			filelist = readdirRecursive(filepath, filelist)
+		} else if (stat.isFile()) {
+			filelist = filelist.push(filepath)
+		}
```
lukechampine, line 85: I think this should be a `reduce` rather than a `forEach`.
All else being equal, avoid `forEach` because it encourages side effects.

plugins/Files/js/sagas/files.js
avahowell: You cannot `yield` from a function that isn't a generator.  This rules out using any higher order functions.

plugins/Files/js/sagas/helpers.js
```diff
@@ -0,0 +1,161 @@
...
+	let parsedDownloads = List(downloads).filter((download) => Date.parse(download.starttime) > since)
+	parsedDownloads = parsedDownloads.map((download) => ({
+		siapath: download.siapath,
+		name: Path.basename(download.siapath),
+		progress: Math.floor((download.received / download.filesize) * 100),
+		destination: download.destination,
+		type: 'download',
+		starttime: download.starttime,
+	}))
+	return parsedDownloads.sortBy((download) => -download.starttime)
```
lukechampine: don't need to reassign the temp variable each time: `return List().filter().map().sortBy()`

ditto for the other functions here

plugins/Files/js/sagas/helpers.js
```diff
@@ -0,0 +1,161 @@
...
+
+	// Take the 36 cheapest hosts from the list
+	storagePrices = storagePrices.sort((p1, p2) => {
+		if (p2.greaterThan(p1)) {
+			return -1
+		}
+		if (p2.lessThan(p1)) {
+			return 1
+		}
+		return 0
```
lukechampine, line 152: There's got to be a more compact way to sort bignums... but a quick google didn't turn up anything :confused:

plugins/Files/js/sagas/helpers.js
```diff
@@ -0,0 +1,161 @@
...
+	}).take(nhosts)
+
+	// Sum and average the storage prices
+	const averagePricePerByteBlockH = storagePrices.reduce((sum, price) => sum.add(price), new BigNumber(0)).dividedBy(4)
```
lukechampine, line 156: why divided by 4?

plugins/Wallet/js/sagas/wallet.js
```diff
@@ -45,6 +45,7 @@ function *walletUnlockSaga(action) {
 		yield siadCall({
 			url: '/wallet/unlock',
 			method: 'POST',
+			timeout: 999999,
```
lukechampine: does this need to be done in the Terminal plugin as well?

plugins/Files/js/sagas/helpers.js
avahowell, line 57: it's because `parsedFiles` should contain a list of unique `files` and `directories` in `path`.  Without using an `Immutable.Map()` and forEach, this function will return multiple duplicate `directories` if two or more files reside inside the same siapath folder.  Basically the Map is used to ensure the returned list is a set of unique files and directories.

plugins/Files/js/reducers/files.js
lukechampine: tabs for indentation, spaces for alignment :neckbeard:

plugins/Files/js/sagas/files.js
lukechampine: interesting. Although this will definitely need to change in the near future, when the call to `/renter` will no longer block.

plugins/Files/js/sagas/files.js
lukechampine: dang, that's annoying. Still, you should be able to loop over the values instead of the index, right? Even if you have to use `.toList`, I think it's cleaner.

plugins/Wallet/js/sagas/wallet.js
avahowell: No.  I was worried that the default timeout would be too small so I tested this- turns out there is no default timeout.  removing this...

plugins/Files/js/sagas/helpers.js
avahowell, line 156: This is the algorithm that David gave me.  I'm not completely sure of all the reasoning behind it

DavidVorick, line 156: You get the average host price, then multiply by the redundancy to get the estimated price. Then you add 50% on top for the cost of downloading files. 1 / 36 \* 6 \* 3 / 2 = 1 / 4

plugins/Files/js/sagas/files.js
avahowell: FYI I changed this to use a map() to construct the actions, and a `for upload in uploads` to dispatch them.

plugins/Files/js/sagas/helpers.js
```diff
@@ -106,16 +106,16 @@ export const totalUsage = (files) => readableFilesize(files.reduce((sum, file) =
...
-		siapath: upload.siapath,
-		name: Path.basename(upload.siapath),
-		progress: Math.floor(upload.uploadprogress),
-		type: 'upload',
-	}))
-	return parsedUploads.sortBy((upload) => upload.name).sortBy((upload) => -upload.progress)
-}
+export const parseUploads = (files) => List(files)
+.filter((file) => file.uploadprogress < 100)
+.map((upload) => ({
```
lukechampine, line 111: indentation here is odd... should probably indent each transform, or maybe move the `List` onto the next line

PR #252, opened 2016-06-20
Fail the build if eslint rules fail

avahowell:
Note that this is failing because there were uncaught linting errors in the terminal plugin merge.


lukechampine:
gonna have to fix those errors before merge :P

PR #253, opened 2016-06-20
Add siafund balance and transfer functionality to the Wallet plugin

avahowell:
This PR adds the ability to send siafunds as well as see your current siafund balance.  A new `SendButton`, `Send Siafunds`, appears if the user has a non-zero balance of siafunds.  A new balance field, `Siafund Balance:`, also appears if the user has a non-zero balance of siafunds.


plugins/Wallet/js/components/wallet.js
```diff
@@ -1,31 +1,37 @@
...
+	const onSendClick = (currencytype) => () => actions.startSendPrompt(currencytype)
+	return (
+		<div className="wallet">
+			<div className="wallet-toolbar">
+				<div className="balance-info">
+					<span>Confirmed Balance: {confirmedbalance} SC </span>
+					<span>Unconfirmed Delta: {unconfirmedbalance} SC </span>
+					{siafundbalance !== '0' ? (<span> Siafund Balance: {siafundbalance} SF </span>) : null}
+				</div>
+				{siafundbalance !== '0' ? <SendButton currencytype="siafunds" onClick={onSendClick('siafunds')} />: null}
```
lukechampine: fyi, I think `if x { <foo></foo> }` is equivalent to `x ? <foo></foo> : null`. Might be cleaner in some cases.

plugins/Wallet/js/reducers/sendprompt.js
```diff
@@ -1,16 +1,19 @@
...
+	currencytype: 'siacoins',
 })
 export default function sendPromptReducer(state = initialState, action) {
 	switch (action.type) {
 	case SET_SEND_AMOUNT:
 		return state.set('sendamount', action.amount)
 	case SET_SEND_ADDRESS:
 		return state.set('sendaddress', action.address)
+	case START_SEND_PROMPT:
+		return state.set('sendtype', action.currencytype)
```
lukechampine: shouldn't this be `currencyType`?
As a whole, I'm not sure I understand why the currencyType is part of the prompt's state, since it's not modifiable (right?). Is it because the other actions need access to it?

plugins/Wallet/js/components/sendbutton.js
```diff
@@ -1,13 +1,14 @@
...
-	return (
-		<div onClick={handleSendButtonClick} className="send-button">
-			<i className="fa fa-paper-plane fa-2x"></i>
-			<span>Send Siacoin</span>
-		</div>
-	)
+const SendButton = ({currencytype, onClick}) => (
+	<div onClick={onClick} className="send-button">
+		<i className="fa fa-paper-plane fa-2x"></i>
+		{currencytype === 'siacoins' ? <span>Send Siacoin</span> : <span>Send Siafunds</span>}
```
lukechampine: not 100% confident, but  I think I'd prefer the type to be passed directly as the text to render, i.e. `<span>Send {currencyType}</span>`. I don't like branching on magic string values.

plugins/Wallet/js/components/wallet.js
avahowell: [if/else doesnt work in JSX](https://facebook.github.io/react/tips/if-else-in-JSX.html)

PR #262, opened 2016-06-21
revise price estimation algorithm and require a minimum of 12 hosts to buy storage

avahowell:


plugins/Files/js/sagas/helpers.js
```diff
@@ -138,22 +138,12 @@ const bytesPerGB = new BigNumber('1000000000')
...
-	}).take(nhosts)
-
-	// Sum and average the storage prices
-	const averagePricePerByteBlockH = storagePrices.reduce((sum, price) => sum.add(price), new BigNumber(0)).dividedBy(4)
+	const minimumHosts = 14
+	const hostPrices = List(hosts).map((host) => new BigNumber(host.storageprice))
+	if (hostPrices.size < minimumHosts) {
+		throw 'not enough hosts'
+	}
+	const averagePricePerByteBlockH = hostPrices.reduce((sum, price) => sum.add(price), new BigNumber(0)).dividedBy(hostPrices.size).times(9)
```
DavidVorick, line 149: should add a comment explaining where the '9' comes from

DavidVorick, line 149: and probably a TODO explaining that this functionality should really be handled by the API, but no endpoint yet exists.

Issue #275, opened 2016-06-22
commands in terminal plugin only return echo + spinning gear


TacoByte:
Darwin 1.0 RC

When I enter a command in the terminal plugin, I get a return of my command + spinning gear, but nothing else happens

avahowell:
Fixed in #276

PR #276, opened 2016-06-22
Check location of siad if siac is not found

droghio:
Restores previous config loading logic and automatically checks the siad enclosing folder for siac before prompting the user. Should correct issue #275.

### DIFF
--- a/plugins/Terminal/js/utils/helpers.js
+++ b/plugins/Terminal/js/utils/helpers.js
@@ -7,7 +7,7 @@ import url from 'url'
 import * as constants from '../constants/helper.js'
 
 export const checkSiaPath = () => new Promise((resolve, reject) => {
-	fs.stat(Path.join(SiaAPI.config.attr('siad').path, process.platform === 'win32' ? '../siac.exe' : '../siac'), (err) => {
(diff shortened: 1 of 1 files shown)


plugins/Terminal/js/utils/helpers.js
```diff
@@ -7,7 +7,7 @@ import url from 'url'
 import * as constants from '../constants/helper.js'
 
 export const checkSiaPath = () => new Promise((resolve, reject) => {
-	fs.stat(Path.join(SiaAPI.config.attr('siad').path, process.platform === 'win32' ? '../siac.exe' : '../siac'), (err) => {
+	fs.stat(Path.resolve(SiaAPI.config.attr('siac').path, './siac'), (err) => {
```
lukechampine, line 10: what about windows?

avahowell, line 10: Isn't `siac.path` the absolute path to siac?  As such, `Path.resolve(siacpath, ./siac)` would return [siacpath]/siac/siac, no?

avahowell, line 10: Looks like `siac.path` is interpreted as the path of the _directory_ containing siac in the terminal plugin.  `siad.path` refers to the absolute path to the siad binary in other parts of the code, so `siac.path` should do the same

PR #278, opened 2016-06-22
fix uploadFolder saga action iteration

avahowell:


plugins/Files/js/sagas/files.js
```diff
@@ -133,8 +133,8 @@ function *uploadFolderSaga(action) {
 			source: file,
 		})).map((file) => actions.uploadFile(file.siapath, file.source))
 
-		for (const upload in uploads) {
-			yield put(uploads[upload])
+		for (const upload in uploads.toArray()) {
+			yield put(uploads.get(upload))
```
lukechampine, line 137: can you iterate over the values of `uploads.toArray()`?

avahowell, line 137: no, the semantics of `for x in y` where `y` is an array mean that x is an index in array `y`.

Issue #277, opened 2016-06-22
Please add more documentation regarding how to configure host and set up storage space for renting


roymj88:
The current documentation do not contain any detail regarding how to configure the storage space and set up price and modify storage space, location etc using sia-ui. Please update or add more info regarding the same.

lukechampine:
A hosting plugin is in the works. It will be included in the 1.0 release.

roymj88:
@lukechampine : Okay cool. Will wait for it.

DavidVorick:
is this issue considered resolved now that there is a hosting plugin available?

lukechampine:
I'll close this for now. If people have specific feedback on changes they'd like to see in the hosting plugin, then they can open a new issue.

PR #281, opened 2016-06-23
Updated Hosting Plugin

droghio:
screen shot 2016-06-22 at 8 44 33 pm

A plugin for file hosting. Allows users to set common host configuration, announce hosts, and configure storage folders. Plugin uses redux and react in a similar style to the new Terminal, Files, and Wallet plugins. Includes simple help menu and information dialogs to make hosting more accessible, as per issue #277. See @DavidVorick's forum post for more info about hosting: http://blog.sia.tech/2016/05/26/how-to-run-a-host-on-sia/


avahowell:
+189,656, lol.  Remove the dist/ folder, I'll add it to the gitignore too

droghio:
@johnathanhowell, what don't think I can type that fast? Made the commit but had to rebase so it might mess with the github pull timeline.

plugins/Hosting/js/components/fileslist.js
```diff
@@ -0,0 +1,53 @@
+import React from 'react'
+import { Map } from 'immutable'
+
+const FilesList = ({ folders, actions }) => {
+	const addStorageLocation = () => actions.addFolderAskPathSize()
+	const removeStorageLocation = (folder) => actions.showWarning(Map({
+		title: 'Delete storage folder?',
+		message: 'No longer use this folder for storage? You may loose collateral if you do not have enough space to fill all contracts.',
```
lukechampine: lose

plugins/Hosting/js/components/resizeDialog.js
```diff
@@ -0,0 +1,43 @@
...
+				<div className="close-button" onClick={closeResizeDialog}>
+					X
+				</div>
+
+				<h3>Resize &quot;{path.basename(resizePath)}&quot;</h3>
+				<p>
+					<label>Size in GB (Min is 35 GB)</label>
+					<input type="number" onChange={handleSettingInput} onKeyDown={handleSettingKeyDown} value={resizeSize} min="35"></input>
+				</p>
+				<span className={'error' + ( resizeSize < 35 ? '' : ' hidden' )}>Storage folder must be at least 35 GB.</span>
```
lukechampine, line 34: @johnathanhowell, how do you feel about using ternary expressions to do styling?

plugins/Hosting/js/reducers/hosting.js
```diff
@@ -0,0 +1,96 @@
...
+		shouldShowResizeDialog: false,
+		resizePath: '',
+		resizeSize: 0,
+		initialSize: 0,
+		warningModalTitle: '',
+		warningModalMessage: '',
+		shouldShowWarningModal: false,
+		shouldShowAnnounceDialog: false,
+		announceAddress: '',
+	}),
```
lukechampine: @johnathanhowell, thoughts on this pattern? I think your approach has been to split these modals into their own component and pass things like `warningModalTitle` as a prop, rather than making it part of the state.

plugins/Hosting/js/reducers/hosting.js
```diff
@@ -0,0 +1,96 @@
...
+
+export default function hostingReducer(state = initialState, action) {
+	switch (action.type) {
+	case constants.UPDATE_SETTING:
+		let settingslist = state.get('usersettings')
+		settingslist = settingslist.map((value) => (
+			(value.get('name') === action.setting) ?
+				value.set('value', action.value)
+			: value
+		))
```
lukechampine: I think immutable.Map has a `merge` function that might be of use here

plugins/Hosting/js/reducers/hosting.js
```diff
@@ -0,0 +1,96 @@
...
+
+	case constants.SHOW_RESIZE_DIALOG:
+		return state.set('modals', state.get('modals')
+			.set('shouldShowResizeDialog', true)
+			.set('resizePath', action.folder.get('path'))
+			.set('resizeSize', action.folder.get('size'))
+			.set('initialSize', action.ignoreInitial ? 0 : action.folder.get('size')))
+
+	case constants.HIDE_ANNOUNCE_DIALOG:
+		return state.set('modals', state.get('modals').set('shouldShowAnnounceDialog', false))
```
lukechampine: `setIn` might be clearer here

plugins/Hosting/js/reducers/hosting.js
```diff
@@ -0,0 +1,96 @@
...
+				? action.data.get('usersettings') : state.get('defaultsettings'))
+			.set('settingsChanged', action.ignoreSettings ? state.get('settingsChanged') : false)
+			.set('acceptingContracts', action.data.get('acceptingContracts'))
+			.set('numContracts', action.data.get('numContracts'))
+			.set('storage', action.data.get('storage'))
+			.set('earned', action.data.get('earned'))
+			.set('expected', action.data.get('expected'))
+			.set('files', action.data.get('files'))
+			.set('walletLocked', action.data.get('walletLocked'))
+			.set('defaultAnnounceAddress', action.data.get('defaultAnnounceAddress'))
```
lukechampine: use `state.withMutations` to make this more performant. `merge` may also be helpful here (`merge` uses `withMutations` internally)

droghio:
Just so I don't forget, from my conversation with @DavidVorick:

Need 2 host UX things
1. There needs to be a big warning if the wallet is locked - host can't form contracts if the wallet is locked
2. There needs to be a big warning if the wallet has less than 50 KS in it - host can't form contracts if the wallet has less money than that.

(DONE)

plugins/Hosting/js/components/resizeDialog.js
avahowell, line 34: I think ternaries for styling is fine.  Looks like this always displays the modal component, but sets `hidden` in css if it shouldn't be shown.  Not sure how this differs from ternarying the component rendering (`{shouldRender ? <Component /> : null`}).  @droghio pointed out that it would allow us to implement css animations, it might be more performant too since there's less dom mounting/unmounting.

plugins/Hosting/js/index.js
```diff
@@ -0,0 +1,28 @@
...
+import { createStore, applyMiddleware } from 'redux'
+import createSagaMiddleware from 'redux-saga'
+import { Provider } from 'react-redux'
+import rootReducer from './reducers/index.js'
+import initSaga from './sagas/saga.js'
+import HostingApp from './components/app.js'
+import * as actions from './actions/actions.js'
+
+
+// Render the wallet plugin
```
avahowell: should be `hosting`

plugins/Hosting/js/index.js
```diff
@@ -0,0 +1,28 @@
+// index.js: main entrypoint for the Sia-UI wallet plugin.
```
avahowell: hosting

plugins/Hosting/js/reducers/hosting.js
```diff
@@ -0,0 +1,96 @@
...
+import * as constants from '../constants/constants.js'
+
+const initialState = Map({
+	numContracts: 0,
+	storage: 0,
+	usersettings: List([
+		Map({ name: 'Max Duration (Weeks)', value: 0, min: 12 }),
+		Map({ name: 'Collateral per TB per Month (SC)', value: 0 }),
+		Map({ name: 'Price per TB per Month (SC)', value: 0 }),
+		Map({ name: 'Bandwidth Price (SC/TB)', value: 0 }),
```
avahowell: What if we had a `settings` reducer that had `maxduration`, `collateral`, `price`, `bandwidthprice`?  Actions would be `setCollateral, setPrice, setBandwidthPrice, setMaxDuration`, etc.   Seems much clearer and less error prone to me

plugins/Hosting/js/reducers/hosting.js
avahowell: Yes, the less that's in the state the better.  My approach with stateful modals (`Buy Storage`, `Send Siacoins`, etc), is to split them into their own reducer.  I definitely think this reducer is too large and should be split up.  Modals should be stateless components that are passed props.

plugins/Hosting/js/reducers/hosting.js
avahowell: I definitely think settings should be held in their own reducer and have their own actions, passing around a big chunk of state with `updateSettings` and then mapping over that chunk seems like we're just implementing a _worse_ reducer that's harder to reason about.

plugins/Hosting/js/sagas/saga.js
```diff
@@ -0,0 +1,233 @@
...
+		url: '/storage',
+		method: 'GET',
+	}).then((fetchedFiles) => {
+		resolve(List(fetchedFiles.StorageFolderMetadata).map((file) => (
+			Map({
+				path: file.path,
+				size: (new BigNumber(file.capacity)).times('1e-9').toString(),
+				free: (new BigNumber(file.capacityremaining)).times('1e-9').toString(),
+			})
+		)).toList())
```
lukechampine: don't bother converting to and from an `immutable.List` within the same expression. Immutable objects are only really helpful if you're passing them around.

plugins/Hosting/js/sagas/saga.js
```diff
@@ -0,0 +1,233 @@
...
+		}
+	} catch (e) {
+		SiaAPI.showError({ title: 'Error Resizing Folder', content: e.message })
+	}
+}
+
+const MAX_DUR = 0
+const COLLATERAL = 1
+const PRICE = 2
+const BANDWIDTH = 3
```
lukechampine: this is definitely smelly.

plugins/Hosting/js/utils/host.js
```diff
@@ -0,0 +1,37 @@
...
+	values.reduce((isValid, val) => (!isNaN(val.val) && val.val > val.min) && isValid, true)
+)
+
+export const hastingsByteBlockToSCTBMonth = (hastings) => (
+	hastingsByteToSCTB(hastings).times('4320') //4320 = blocks per month
+)
+
+export const SCTBMonthToHastingsByteBlock = (SC) => (
+	SCTBToHastingsByte(SC).dividedBy('4320') //4320 = blocks per month
+)
```
lukechampine, line 34: @johnathanhowell, I think these are worth adding to Nodejs-Sia

package.json
```diff
@@ -38,7 +38,8 @@
   },
   "scripts": {
     "start": "webpack && electron .",
-    "build": "webpack",
+    "debug": "webpack -d && electron .",
+    "build": "webpack -d",
```
avahowell: `build` should not make sourcemaps, only `debug`

out.txt
```diff
@@ -0,0 +1 @@
+ornament issued surfer shyness square energy rewind zesty iceberg aplomb wade dual pistons eight ambush village legion badge southern upper tweezers calamity dogs ornament mundane athlete shuffled soda acidic
```
lukechampine: be more careful with what you stage. `git add *` is always a bad idea.

droghio: It's a bogus seed at least. I think a few files snuck in when I rebased. To the git time machine.....

lukechampine:
Looks good, my only remaining gripes are:
- Changing the `Accepting Contracts` setting does not seem to activate the "Save" button.
- If you announce without any money, you get an ugly error message (I assume piped straight from the API)

PR #292, opened 2016-06-24
Files plugin tweaks

avahowell:
this PR makes `uploads` come after `downloads` in the file transfers list, changes styling to use the correct pointer types, and fixes a typo.

### DIFF
--- a/plugins/Files/js/components/filetransfers.js
+++ b/plugins/Files/js/components/filetransfers.js
@@ -12,8 +12,8 @@ const FileTransfers = ({uploads, downloads, actions}) => {
 			<div className="close-button" onClick={onCloseClick}>
 				<i className="fa fa-times fa-2x"></i>
 			</div>
-			{uploads.size > 0 ? <UploadList uploads={uploads} /> : null}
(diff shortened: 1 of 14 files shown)


plugins/Files/js/components/filetransfers.js
```diff
@@ -12,8 +12,8 @@ const FileTransfers = ({uploads, downloads, actions}) => {
 			<div className="close-button" onClick={onCloseClick}>
 				<i className="fa fa-times fa-2x"></i>
 			</div>
-			{uploads.size > 0 ? <UploadList uploads={uploads} /> : null}
 			{downloads.size > 0 ? <DownloadList downloads={downloads} onDownloadClick={onDownloadClick} /> : null}
+			{uploads.size > 0 ? <UploadList uploads={uploads} /> : null}
```
lukechampine, line 16: doesn't `UploadList` need to be passed a `onUploadClick`?

avahowell, line 16: no, `onUploadClick` gets passed to transfer list's `onTransferClick`, which has an es6 default parameter value.  If you pass undefined, it's the same as passing `() => {}`

avahowell:
dont merge yet, found a bug

avahowell:
Should be ready now

PR #293, opened 2016-06-25
Separate 'no matching files' and 'no files uploaded' text

avahowell:
This pr changes filelist to return 'no matching files' when no search results are returned, and 'no files uploaded' when no files have been uploaded.


plugins/Files/js/sagas/helpers.js
```diff
@@ -119,6 +119,7 @@ export const parseUploads = (files) => List(files)
 
 // Search `files` for `text`, excluding directories not in `path`
 export const searchFiles = (files, text, path) => {
+	console.log(text)
```
lukechampine: purge

avahowell: ack, amended and force pushed

PR #304, opened 2016-06-27
Add storage usage feedback

### LINKED ISSUE #298: I can't see how much storage I bought
After buying 10 GB of space on the network, there is no way for me to check that stat in the UI. It's nice to know how much I used, but what about how much I have left?

It would be really useful/informative to know how much space I have used, available, and total. Some sort of billing history would be nice as well.

avahowell:
This PR adds a `UsageStats` component which shows the user how much storage they have used and how much they have available.  `Available Storage` is estimated by using the average host price algorithm and computing the amount of storage the user's `allowance` could purchase.

Should close #298


plugins/Files/js/sagas/helpers.js
```diff
@@ -134,19 +134,33 @@ export const searchFiles = (files, text, path) => {
...
-	const averagePricePerByteBlockH = hostPrices.reduce((sum, price) => sum.add(price), new BigNumber(0)).dividedBy(hostPrices.size).times(9)
+	return hostPrices.reduce((sum, price) => sum.add(price), new BigNumber(0)).dividedBy(hostPrices.size).times(9)
+}
+
+// Take an allowance and return a number of bytes this allowance
+// can be used to store, given a list of hosts to store data on.
+export const allowanceStorage = (funds, hosts, period) => {
+	const allowanceFunds = new BigNumber(funds)
+	const costPerB = estimatedStoragePriceH(hosts).times(period)
+	return readableFilesize(allowanceFunds.dividedBy(costPerB).toNumber())
```
lukechampine: there's probably a divide-by-zero bug in here

Issue #296, opened 2016-06-25
Upgrading is messy, confusing, and a bit scary


TacoByte:
RC2 Darwin

I imported the consensus and wallet files before starting the RC2.

It hung for a few minutes before managing to boot into the main screen. Now it's been stuck on "Loading Sia". I guess I'll just wait a bit longer? Peeking inside the Sia folder, I can tell thing are happening as some files' last modified timestamps match the current time. I'm assuming things are going in the right direction and the program is not just stuck in a loop... Or is it? No way to tell for sure.

There needs to be more feedback for the user. I can't determine whether Sia has crashed or what is happening. In-app upgrading is also needed, messing with wallets and other files is not nice and potentially a bit dangerous if you're careless or easily distracted.

TacoByte:
It's slowly calculating the usual folders and files. transactionpool, miner, gateway, host, renter etc... Not sure why so long since I'm not doing any renting/mining/farming. Right now it looks like it's going through the whole blockchain and refreshing contractor.json as it reads the contracts. Perhaps I should have moved the rest of the folders/files as well?

I'm even wondering if this might have an impact on uptime calculations as the whole process is probably going to clock in at around 1 hour. If this is going to be the case for every upgrade, doesn't that leave significantly less time for other kinds of planned and unplanned downtime? If you upgrade twice a month (not unusual), that's a whole day in the year spent upgrading!

If at all possible, I would suggest this to be a 1.0 requisite (proper feedback + in-app upgrading)

avahowell:
Yeah, `siad` can take a long time to rescan an incomplete data-set.  The UI shows "Loading..." when siad has been started but is not finished loading modules, "Welcome Back" when it finishes, and an error if it fails.  There's a planned upgrade to load the API module first, which means the UI can provide more granular feedback on this process.

As for upgrading, the plan is to use a common, OS-specific, data directory so users can just replace their binary and not have to worry about meddling with Sia data.  That issue is here: https://github.com/NebulousLabs/Sia-UI/issues/243 .  It should be implemented for 1.0.  In-app upgrades are much harder, but also planned https://github.com/NebulousLabs/Sia-UI/issues/255.

Thanks again for your feedback.

TacoByte:
Thank you for listening!

[...]

Issue #243, opened 2016-06-07
Use an OS-standard location for siad's `--sia-directory`


avahowell:
An OSX user recently pointed out that since the default Sia `datadir` is located in the App bundle, there's no way of upgrading their UI release without wiping out Sia's data.  Since a recent pull request to `sia.js`, it's now possible to set `--sia-directory` independently from `siad`'s path.  Using an OS-standard data location is the obvious solution to this problem (`ApplicationSupport` for OSX, `%APPDATA%` for windows, `~/.local/share/sia` for linux), but there are caveats with this as discussed in https://github.com/NebulousLabs/Sia/issues/1070.  In order to safely implement this change, `siad` should add a lockfile to the `sia-directory` to ensure only one instance of `siad` can be running on a given directory.

In the mean time, we can guide users through the process of migrating their data from the app bundle to the new location upon 1.0's release.  This process is easier for Windows and Linux users since they can directly replace the binary inside their release folder, unlike OSX users.

DavidVorick:
There might be a way to do in-place upgrades from within the app, which would be the preferred method. Using a standard data directory has previously been associated with losing money. Now that we have seed based wallets it's much less of a risk, but still an issue if we end up with multiple instances of `siad` open at the same time.

You can use things like file locks to help protect against this problem, but if we start using a common directory we also lose portability and we lose the ability for a user to easily have multiple wallets on the same machine.

VoidingWarranties:
Closed in #306.

PR #306, opened 2016-06-27
use electron userData path as the default sia data directory

avahowell:
Implements #243.  Makes #296 easier, since users no longer have to manually migrate their data.

### DIFF
--- a/js/mainjs/config.js
+++ b/js/mainjs/config.js
@@ -1,12 +1,13 @@
 import fs from 'fs'
 import Path from 'path'
+import { app } from 'electron'
 
 // The default settings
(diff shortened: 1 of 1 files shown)


lukechampine:
we need to make users aware of this change (not a popup in the UI or anything, just documentation and such alongside the release). Also, can you try running two instances of Sia-UI and see what happens?

avahowell:
Running two instances of Sia-UI causes the second instance to attach to the first's running instance of `siad`.  Attempting to run a new instance of siad with new ports (api, host, and rpc) causes the consensus module to timeout.

PR #309, opened 2016-06-28
update UI to use new hostdb route and new error json

avahowell:


lukechampine:
what about the host? Is drogo handling that?

avahowell:
Host is already doing it https://github.com/NebulousLabs/Sia-UI/blob/master/plugins/Hosting/js/sagas/saga.js#L142 

> e.message

PR #312, opened 2016-06-28
add mocha, chai, sinon, enzyme testing stack, FileList unit tests

avahowell:
This PR adds the necessary dependencies for testing the UI.  Included is the testing for the FileList component of the renter, more tests will follow in later PRs.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["es2015", "react"]
+}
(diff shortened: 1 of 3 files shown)


.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["es2015", "react"]
+}
```
lukechampine, line 3: damn, I thought we could get away with not having a babelrc

lukechampine:
From Travis:

```
Warning: Failed propType: Required prop `showSearchField` was not specified in `FileList`.
```

Should we bother correcting these?

avahowell:
Yeah, I think a required prop should always be passed even in testing.

.babelrc
avahowell, line 3: `--compilers js:babel-register` :/

PR #314, opened 2016-06-28
wallet component unit tests

avahowell:
This PR adds unit tests for the wallet component.

### DIFF
--- a/test/wallet/wallet.component.js
+++ b/test/wallet/wallet.component.js
@@ -21,10 +21,13 @@ describe('wallet component', () => {
 	it('should render balance info', () => {
 		const walletComponent = shallow(<Wallet confirmedbalance="10" unconfirmedbalance="1" siafundbalance="0" />)
 		expect(walletComponent.find('.balance-info').children()).to.have.length(2)
+		expect(walletComponent.find('.balance-info').children().first().text()).to.contain('Confirmed Balance: 10 SC')
+		expect(walletComponent.find('.balance-info').children().last().text()).to.contain('Unconfirmed Delta: 1 SC')
 	})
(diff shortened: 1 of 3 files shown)


test/wallet/wallet.component.js
```diff
@@ -0,0 +1,78 @@
...
+		const walletComponent = shallow(<Wallet showSendPrompt={true} confirmedbalance="10" unconfirmedbalance="1" siafundbalance="0" actions={testActions} />)
+		expect(walletComponent.contains(<SendPrompt />)).to.equal(true)
+	})
+	it('should not show receive prompt unless showReceivePrompt', () => {
+		const walletComponent = shallow(<Wallet showReceivePrompt={false} confirmedbalance="10" unconfirmedbalance="1" siafundbalance="0" actions={testActions} />)
+		expect(walletComponent.contains(<ReceivePrompt />)).to.equal(false)
+	})
+	it('should show receive prompt when showReceivePrompt', () => {
+		const walletComponent = shallow(<Wallet showReceivePrompt={true} confirmedbalance="10" unconfirmedbalance="1" siafundbalance="0" actions={testActions} />)
+		expect(walletComponent.contains(<ReceivePrompt />)).to.equal(true)
```
lukechampine: I believe you can use `.to.be.true` as shorthand here.
There's also the `should` style, which (AFAICT) would look like:

``` js
walletComponent.contains(<ReceivePrompt />).should.be.true
```

avahowell: `should` looks like this:

```
import { should } from 'chai'
walletComponent.contains(<ReceivePrompt />).should().be.true
```

I think I like expect() style better.

PR #315, opened 2016-06-28
change default data directory to userData/sia-data

avahowell:


lukechampine:
why not just `sia`?
I assume that without this, you wind up with `gateway`, `consensus`, etc. folders in your `.config`?

avahowell:
`userData` is located in `~/.config/Sia-UI-Beta`.  I like the idea of having the sia data in a discrete directory.  I'll change it to `sia`

PR #337, opened 2016-06-29
only allow one instance of Sia-UI

avahowell:


lukechampine:
What is the effect of this change?

avahowell:
If there's already an instance of the UI running and the user launches the UI binary, the running instance's window will restored and focused.  Previously, a new instance would be spawned.

PR #339, opened 2016-06-29
use a filtered average to calculate sane renter price

avahowell:
In the event of outlier hosts with very high prices, the price estimation algorithm returns very inflated prices.  This PR changes that algorithm to filter outlier prices before computing the average.

### DIFF
--- a/plugins/Files/js/sagas/helpers.js
+++ b/plugins/Files/js/sagas/helpers.js
@@ -149,7 +149,7 @@ export const estimatedStoragePriceH = (hosts) => {
 	}
 
 	// Compute the average host price.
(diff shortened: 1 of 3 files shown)


plugins/Files/js/sagas/helpers.js
```diff
@@ -149,11 +149,13 @@ export const estimatedStoragePriceH = (hosts) => {
 	}
 
 	// Compute the average host price.
-	// Multiply this average by 9 to assume a redundancy of 6 with 25% of the files being downloaded at 2x monthly price
+	// Multiply this median by 9 to assume a redundancy of 6 with 25% of the files being downloaded at 2x monthly price
```
lukechampine: comment is wrong

avahowell: ack

Issue #329, opened 2016-06-29
Add an indicator on the wallet page if the blockchain is still being downloaded


AdamBLevine:
A normal user downloading and installing Sia-UI will generally want to send themselves some SIA from an exchange to get started.  Until the blockchain has been downloaded, their wallet will not see their transaction and it's very confusing trying to figure out why your funds are missing.  This is exacerbated because every time you click "Receive" it generates a unique payment address, so the user can not know if they've mis-copied or made some other mistake when in reality the wallet is just still downloading the blockchain.

Add some kind of indication or notification to the wallet page alerting users that transactions won't be detected by the wallet until they've finished downloading the blockchain.

avahowell:
A status indicator in the top bar would probably be helpful for this, along with a notification in the wallet plugin.

AdamBLevine:
That would do it, thanks for the quick response!

avahowell:
Closed with #349

PR #349, opened 2016-07-04
Add sync status indicator to the topbar

avahowell:
This PR improves UX by adding a blockchain sync status indicator to the top bar of the UI, displayed in every plugin.  Previously users would be confused why coins were not shown in their wallet, due to their wallet not being synchronized.

Addresses #329.

screen shot 2016-07-04 at 6 37 50 pm
screen shot 2016-07-04 at 6 38 17 pm


lukechampine:
neat!

Issue #341, opened 2016-06-30
Cannot Upload Individual files in Linux/Windows


droghio:
Discovery of this issue goes to thapgich in the Slack. On Linux and Windows you cannot upload files, only folders. This has to due to a limitation of Electron. From http://electron.atom.io/docs/api/dialog/:

**Note: On Windows and Linux an open dialog can not be both a file selector and a directory selector,  so if you set properties to ['openFile', 'openDirectory'] on these platforms, a directory selector will be  shown.**

Fornax96:
![image](https://cloud.githubusercontent.com/assets/1247928/16502373/588cf360-3f0e-11e6-9dca-ce5588e41bd5.png)

EDIT: I also can't multi select (With shift or ctrl pressed). This needs some work

avahowell:
@droghio has a pr that should come in soon that fixes this issue.  There will be two buttons that drop down from 'Upload', 'Upload Files' and 'Upload Folder'.  For now, you can drag and drop files into the renter to upload them.

avahowell:
closed via #350

PR #350, opened 2016-07-05
Add individual file upload on Windows and Linux

droghio:
Fixes issue #341.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "jquery": "^2.1.4",
     "json-loader": "^0.5.4",
     "mkdirp": "^0.5.1",
+    "proxyquire": "^1.7.10",
     "react": "^15.0.2",
     "react-dom": "^15.0.2",
     "react-redux": "^4.4.5",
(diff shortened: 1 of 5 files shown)


avahowell:
👍

tests would be helpful for this, did you test the behavior manually?  We can mock the `fs` calls using [rewire](https://github.com/jhnns/rewire), mocking `SiaAPI` is as simple as defining a `global.SiaAPI` in the test script.

droghio:
I tested the behavior on Windows/Mac up until it sends the post call to the api. I might setup a test net to confirm the uploads are working as expected. I think we might be able to just get ride of the folder upload button all together, as long as people can select multiple files/drag and drop. I'll work on tests in the mean time.

avahowell:
You don't need to test the entire upload process, only that this component has the expected behaviour (it calls uploadFolder with the selected folder(s) (if it's a folder), uploadFile with the selected file(s), dispatches the expected actions, etc)

avahowell:
LGTM

Issue #348, opened 2016-07-04
Terminal plugin doesn't clear password or wallet seed fields


VoidingWarranties:
Repro steps:
1. `wallet load seed`
2. type in password
3. type in seed
4. `wallet load seed`

Both the password and seed fields are still filled in with the previous input.

PR #353, opened 2016-07-05
Clear wallet passwords on submit

droghio:
Fixes issue #348. Updates help menu with new siac commands.

### DIFF
--- a/plugins/Terminal/js/components/seedprompt.js
+++ b/plugins/Terminal/js/components/seedprompt.js
@@ -22,6 +22,8 @@ export default class WalletSeedPrompt extends React.Component {
 					'dictionary': 'english',
 				}))
 				siac.end()
+				e.target.value = ''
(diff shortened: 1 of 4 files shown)


plugins/Terminal/js/components/seedprompt.js
```diff
@@ -22,6 +22,8 @@ export default class WalletSeedPrompt extends React.Component {
 					'dictionary': 'english',
 				}))
 				siac.end()
+				e.target.value = ''
```
avahowell, line 25: Is `e.target.value =''` necessary?  The input's value isn't stored as state, so it should be cleared every time the seed prompt is reopened.

droghio, line 25: Since I am only adding or removing class attributes instead of changing the dom the value is stored unless I manually clear it.

PR #355, opened 2016-07-06
Gracefully disable plugins if Siad is stopped or crashes

avahowell:
This PR implements changes to the plugin API that ensures all plugins are unloaded if Siad becomes unreachable, for example if the user runs 'stop' in the terminal plugin, or if Siad crashes.  Previously, a connection error would be thrown every time a plugin polled the API.

screen shot 2016-07-05 at 7 48 52 pm
screen shot 2016-07-05 at 7 49 06 pm


avahowell:
@lukechampine, hold off merging this until I've added some tests.  I want every new PR to have unit tests.

lukechampine:
ok, but in that case you should drop a new comment when you've finished adding tests :octocat:

avahowell:
ACK, should be good to go

js/rendererjs/disabledplugin.js
```diff
@@ -0,0 +1,24 @@
...
+
+const containerStyle = {
+	display: 'flex',
+	alignItems: 'center',
+	justifyContent: 'center',
+	flexDirection: 'column',
+	backgroundColor: '#C6C6C6',
+	width: '100%',
+	height: '100%',
+}
```
lukechampine, line 11: some inline CSS, eh? Is there a good metric for when to use this and when to define the style in a .css file?

js/rendererjs/pluginapi.js
```diff
@@ -2,13 +2,52 @@
...
 Siad.configure(config.siad)
+let disabled = false
+
+window.onload = () => {
+	// ReactDOM needs a DOM in order to be imported,
+	// but the DOM is not available until the plugin has loaded.
+	// therefore, we have to global require it inside the window.onload event.
+
+	/* eslint-disable global-require */
+	const ReactDOM = require('react-dom')
```
lukechampine, line 19: +1 for explanatory comment

js/rendererjs/disabledplugin.js
avahowell, line 11: I used inline css here because this component is mounted directly in the `body` of the plugins, and should be styled independently of the plugin's styling.  Adding these css rules to `general.css` doesn't work either, since this component is mounted inside a `<webview>`.  The alternative would be to add an identical css rule to each plugin, or add this rule to `plugin-standard`

js/rendererjs/pluginapi.js
```diff
@@ -2,13 +2,52 @@
...
+	/* eslint-enable global-require */
+
+	const checkSiad = () => {
+		Siad.ifRunning(() => {
+			if (disabled) {
+				disabled = false
+				window.location.reload()
```
lukechampine, line 31: what does this do?

test/pluginapi.unit.js
```diff
@@ -0,0 +1,77 @@
...
+	}
+}
+
+proxyquire('../js/rendererjs/pluginapi.js', mock)
+
+describe('plugin API', () => {
+	it('creates a SiaAPI window object', () => {
+		expect(window.SiaAPI).to.exist
+	})
+	it('doesnt mount disabled plugin component if siad is running', function(done) {
```
lukechampine: typo

lukechampine:
broadly, it seems odd that the siad polling is happening outside the component. Normally you'd have a component with a `siadRunning` boolean prop that would display the overlay if `false` and the plugin if `true`. But I suppose that may not be feasible given the current plugin architecture?

js/rendererjs/pluginapi.js
avahowell, line 31: This polls siad continuously (every 2s), and disables the plugin if siad is not running.  The plugin is 'disabled' by mounting `<DisabledPlugin />` in the DOM's body.  If siad is found to be running, and the plugin is disabled, the window is reloaded, clearing the <DisabledPlugin /> component and reloading the plugin.

avahowell:
It's feasible, but that seems like behaviour we would want in every plugin by default, hence why it's in `pluginapi.js` (`pluginapi.js` gets `preloaded` into each plugin's webview.)

PR #358, opened 2016-07-06
Add a 'Stop Siad' button to the UI

avahowell:
screen shot 2016-07-06 at 6 17 19 pm

### DIFF
--- a/css/general.css
+++ b/css/general.css
@@ -61,6 +61,34 @@ html,body {
 .logo-container:hover {
 	cursor: pointer;
 }
+.stop-button:hover {
+	opacity: 0.75;
+}
+.stop-button span {
+	color: #39383A;
+	margin-left: 5px;
+	margin-right: 5px;
+}
+.stop-button i {
+	color: #E0000B;
+}
+.stop-button {
+	cursor: pointer;
+}
+#siad-sidebar-button {
+	position: absolute;
+	bottom: 0;
+	left: 0;
+	height: 40px;
+	width: 100%;
+	background-color: #C4C0C8;
+	display: flex;
+	align-items: center;
+	justify-content: center;
+	margin: 0;
+	padding: 0;
+	z-index: 1;
+}
 .status-bar div {
 	display: flex;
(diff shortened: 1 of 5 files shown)


DavidVorick:
LGTM

lukechampine:
Do we need to feature the button so prominently? I feel like stopping siad is a pretty uncommon use case. You can stop siad from the Terminal, right?

DavidVorick:
I've had multiple users requests a stop button. Do you feel that the stop button detracts from the overall look and feel / user experience?

lukechampine:
Did they explain why they want it? What is the root problem that this addresses? Are there better ways we can solve this problem?
I think for most users, it's just a confusing new button. How many UI users really understand the relationship between the UI and siad? Many client-daemon apps don't expose that. Does any torrent client have a button that stops the daemon, but leaves the UI open? It just seems weird to me.

droghio:
Is there anyway to restart siad with the default ui preferences once it stops?

avahowell:
@droghio the current behavior (introduced by #355) is to start sia-ui with the current UI preferences (`config.js`)

DavidVorick:
@lukechampine some users want to close the GUI but leave 'siad' running, while others want siad to shutdown when they close the GUI. We've had a lot of trouble finding the right balance between the two.

WaveringAna:
You can have minimize be the close while keeping siad open and the exit as close both

avahowell:
@aayanl this is actually already the case; closing the UI minimizes it to your system tray.  You can use the system tray to exit.

DavidVorick:
@johnathanhowell are you planning on getting this PR ready to merge?

avahowell:
I don't think a `Stop Siad` button really makes sense.  There should only be a quit button, and that already exists in the system tray.  I think a 'Stop Siad' button exposes implementation details to the user that they shouldn't care about.

DavidVorick:
should this PR be closed then?

avahowell:
yes.

PR #360, opened 2016-07-07
change statusbar to show synchronizing, not synchronizing, synchronized

avahowell:
Previoulsy "Not Synchronized" would be shown if `synced` was false, now the UI will show "Not Synchronizing" if there are no peers and consensus is not synced, "Synchronizing" if there are peers and consensus is not synced, and "Synchronized" if consensus is synced.

screen shot 2016-07-07 at 3 48 50 pm
screen shot 2016-07-07 at 3 52 35 pm
screen shot 2016-07-07 at 3 55 36 pm

### DIFF
--- a/css/general.css
+++ b/css/general.css
@@ -69,7 +69,7 @@ html,body {
 	margin-left: 5px;
 	margin-right: 5px;
 }
-.status-bar-unsynced i, .status-bar-synced i {
+.status-bar i {
 	margin-right: 10px;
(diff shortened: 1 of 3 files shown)


lukechampine:
Would be nice to have some indication of the number of peers. Doesn't need to be blatant, but maybe stick it in a tooltip or something. If you can't think of a good approach, just leave it out for now.

PR #361, opened 2016-07-07
build only the master branch

lukechampine:
this allows us to build both PR commits and master branch merges, without double-building each PR push.


avahowell:
👍

PR #362, opened 2016-07-07
Wallet integration tests

avahowell:
This PR introduces end-to-end integration testing for the Wallet plugin.  The entire plugin is initialized and mounted in the `jsdom` virtual DOM, using `enzyme`'s `mount` function.  The Siad API is mocked using a Sinon `stub`.


lukechampine:
not pretty... but it gets the job done

PR #365, opened 2016-07-11
Don't convert siafunds from hastings

### LINKED ISSUE #364: Sia UI does not display some information
Sia UI wallet page does not display transactions of Siafund.

pasted image at 2016_07_11 11_18 am

avahowell:
`parseRawTransactions` should not treat Siafunds as hastings, doing so results in their value being rounded down to zero.  Fixes #364


DavidVorick:
LGTM

PR #367, opened 2016-07-12
dont allow users to select status bar text

avahowell:


DavidVorick:
why don't we want users to be able to select that text?

avahowell:
The status bar is a purely presentational component, I dont think it makes sense to have the text have the "I" cursor icon and be selectable, as if it were a text field.  It's a very minor thing, though.

lukechampine:
I don't think this prevents selection; it just changes the cursor. Which is probably the best behavior. There's certainly some utility in being able to select and copy the blockheight.

PR #373, opened 2016-07-20
ensure every `watch*` saga is run inside the root saga of wallet/files

avahowell:
This further ensures that every exported saga will be run inside the root saga, preventing bugs similar to #369 from occurring in the future.

### DIFF
--- a/plugins/Files/js/sagas/index.js
+++ b/plugins/Files/js/sagas/index.js
@@ -2,20 +2,6 @@ import * as sagas from './files.js'
 import { fork } from 'redux-saga/effects'
 
 export default function* rootSaga() {
-	yield [
-		fork(sagas.watchGetWalletLockstate),
-		fork(sagas.watchGetFiles),
-		fork(sagas.watchSetAllowance),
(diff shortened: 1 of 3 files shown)


plugins/Files/js/sagas/index.js
```diff
@@ -2,20 +2,6 @@ import * as sagas from './files.js'
...
-		fork(sagas.watchUploadFile),
-		fork(sagas.watchDownloadFile),
-		fork(sagas.watchGetDownloads),
-		fork(sagas.watchGetUploads),
-		fork(sagas.watchDeleteFile),
-		fork(sagas.watchUploadFolder),
-		fork(sagas.watchGetStorageMetrics),
-		fork(sagas.watchGetContractCount),
-	]
+	const watchers = Object.values(sagas).map((saga) => fork(saga))
```
lukechampine: is `map(fork)` legal?

avahowell: Yes, the tests verify that this works.  I've manually tested it too.

Issue #371, opened 2016-07-20
Add window icon


Fornax96:
Currently the window is shown as a big question mark on my application dock: 
![image](https://cloud.githubusercontent.com/assets/1247928/16982044/0c59a744-4e6e-11e6-91b6-e7f8081605bf.png) 

This can be confusing since it's not easily distinguished from other apps that don't have a window icon. 

I thought I would be able to fix this myself and started messing around with this line: https://github.com/NebulousLabs/Sia-UI/blob/master/js/mainjs/initWindow.js#L17, but I couldn't get it to work, it always shows the question mark. This makes me think the problem lies a bit deeper than just the icon path.

DavidVorick:
I believe that you need to use a specific Windows-only program to get icons added to Windows binaries. It was a part of our release process at one point, but may have gotten left out at some point as we iterated on our build design. Shouldn't be too hard to fix.

droghio:
@Fornax96 Thanks for the issue report. Just to double check which OS are you running. Your picture looks like a GNOME dock to me.

Fornax96:
It's Ubuntu 16.04 Unity desktop

DavidVorick:
:< hmm ok

I should have known that

droghio:
Looks like this error has been patched an a recent version of Electron: https://github.com/electron/electron/pull/6323
I will double check that was the problem before closing the issue.

droghio:
So the missing icon is caused by two bugs:
1. Due to https://github.com/electron/electron/issues/6205 for Ubuntu/Linux support the icon must be called `icon.png` and be 128x128px if we want to keep using electron 1.2.2. I believe there were bugs in the electron 1.2.3 build which is why we held back.
2. Once we build the js into a release the relative pathing changes since we use web-pack. The icon path needed to be updated.

I have a pull request ready for the path update and a few other minor UI changes but the question is whether we should make the default icon size 128x128px or leave it as 64x64px. I'm not sure if that size is necessary to build the Mac icon. We also need to decide whether to update the electron dependency to `>1.2.2` or leave it as is for now.

@johnathanhowell Any thoughts?

droghio:
Fixed in PR #374

PR #374, opened 2016-07-21
Update Electron Dependency and Fix Linux Icons

droghio:
This fixes a few UI bugs including the lack of an icon under Linux ie #371.


avahowell:
Can you manually verify that electron successfully launches when you run `npm start`?

droghio:
I can do that and one better. Works with npm start, and I also made releases for Mac, Linux, and Windows and made sure they all launch.

avahowell:
perfect, just didn't want to have a similar 'test-pass-but-electron-is-broken' scenario as the last time we updated electron.

LGTM

package.json
```diff
@@ -58,7 +58,7 @@
     "sia-import": "(test -d Sia || mkdir -p Sia) && (cp $GOPATH/bin/siad Sia & cp $GOPATH/bin/siac Sia)",
     "sia-kill": "killall siad; while pgrep siad; do sleep .5; done;",
     "sia-refresh": "npm run sia-kill && rm -rf Sia && npm run sia-import",
-    "release-args": "webpack && electron-packager . Sia-UI --app-version=1.0.1 --version=1.2.2 --overwrite --prune --ignore='(node_modules|release|config.json|errors.log)' --out=release/v1.0.1 --icon=assets/icon",
+    "release-args": "webpack && electron-packager . Sia-UI --app-version=1.0.2 --overwrite --prune --ignore='(node_modules|release|config.json|errors.log)' --out=release/v1.0.2 --icon=assets/icon",
```
lukechampine, line 61: what is the effect of removing the `--version` arg? Will it just use the most recent version?

droghio:
Not exactly, from the electron-packager docs:

> The Electron version with which the app is built (without the leading 'v') - for example, 0.33.9. See 
> Electron releases for valid versions. If omitted, it will use the version of the nearest local installation 
> of electron-prebuilt.

It will default the version of electron prebuilt installed on the system, which is currently defined as `^1.2.8` == `>=1.2.8 < 2.0.0`.

Issue #255, opened 2016-06-21
Enable in-place automatic updates


avahowell:

DavidVorick:
It was mentioned that in-place updates would be scary. Is this something we want to shelve?

Something that might be sufficient - update `siad` automatically in-place, and then direct the user to a location where they can download the most recent UI. We can store the siad and siac binaries in the common dir (not sure if that's already what's happening)

avahowell:
Siad and siac are currently distributed and stored with the UI release binaries- not in the common dir.  I think it's sufficient to have a simple check in the UI for the new version and point users to the download location, since upgrading is trivial with the addition of the common data directory.  I'll keep this issue open until that's implemented.

avahowell:
With our new UI release scripts, the UI releases are now signed, meaning true automatic updating can now be implemented.

PR #377, opened 2016-07-21
check for updates using github on launch and every 30 minutes, display a notification with a link to the new version

avahowell:
This PR adds a floating notification in the bottom right of the UI that appears if the user's UI is out of date with the latest release on github.  Clicking on this notification guides the user to the download page.

Addresses #255

### DIFF
--- a/js/rendererjs/loadingScreen.js
+++ b/js/rendererjs/loadingScreen.js
@@ -25,16 +25,16 @@ const showError = (error) => {
 
 // startUI starts a Sia UI instance using the given welcome message.
 // calls initUI() after displaying a welcome message.
-const startUI = async (welcomeMsg, initUI) => {
+const startUI = (welcomeMsg, initUI) => {
 	// Display a welcome message, then initialize the ui
 	overlayText.innerHTML = welcomeMsg
 
 	// construct the autoupdater component and check for updates every 30 minutes.
 	const checkUpdate = async () => ReactDOM.render(await Update(currentVersion), document.getElementById('update-status-container'))
-	setInterval(async () => {
-		await checkUpdate()
+	setInterval(() => {
+		checkUpdate()
 	}, 18000000)
-	await checkUpdate()
+	checkUpdate()
 
 	// Construct the status bar component and poll for updates from Siad
 	const updateSyncStatus = async function() {
--- a/js/rendererjs/loadingScreen.js
+++ b/js/rendererjs/loadingScreen.js
@@ -29,8 +29,12 @@ const startUI = async (welcomeMsg, initUI) => {
 	// Display a welcome message, then initialize the ui
 	overlayText.innerHTML = welcomeMsg
 
-	// construct the autoupdater component and check for updates
-	ReactDOM.render(await Update(currentVersion), document.getElementById('update-status-container'))
+	// construct the autoupdater component and check for updates every 30 minutes.
+	const checkUpdate = async () => ReactDOM.render(await Update(currentVersion), document.getElementById('update-status-container'))
+	setInterval(async () => {
+		await checkUpdate()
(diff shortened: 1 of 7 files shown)


js/rendererjs/update.js
```diff
@@ -0,0 +1,47 @@
...
+import semver from 'semver'
+import { shell } from 'electron'
+import UpdateComponent from './updatecomponent.js'
+
+const getReleases = (url) => new Promise((resolve, reject) => {
+	request({
+		url,
+		json: true,
+		headers: {
+			'User-Agent': 'Sia-UI',
```
lukechampine, line 12: ?

avahowell, line 12: the github API requires that you set the user agent for all requests, otherwise it returns `403 Forbidden`

js/rendererjs/loadingScreen.js
```diff
@@ -36,10 +37,17 @@ const checkSiaPath = () => new Promise((resolve, reject) => {
...
+const startUI = async (welcomeMsg, initUI) => {
 	// Display a welcome message, then initialize the ui
 	overlayText.innerHTML = welcomeMsg
 
+	// construct the autoupdater component and check for updates every 30 minutes.
+	const checkUpdate = async () => ReactDOM.render(await Update(currentVersion), document.getElementById('update-status-container'))
+	setInterval(async () => {
+		await checkUpdate()
+	}, 18000000)
+	await checkUpdate()
```
lukechampine: does this block? What happens if you don't have internet or the call takes a long time?

avahowell: it does block.  If you don't have an internet connection the UI doesnt load, fixing now...

avahowell: fixed, manually tested as well

js/rendererjs/update.js
lukechampine, line 12: weird. Go sets the user agent by default so I figured `request` would do the same.

js/rendererjs/loadingScreen.js
```diff
@@ -37,16 +37,16 @@ const checkSiaPath = () => new Promise((resolve, reject) => {
...
+const startUI = (welcomeMsg, initUI) => {
 	// Display a welcome message, then initialize the ui
 	overlayText.innerHTML = welcomeMsg
 
 	// construct the autoupdater component and check for updates every 30 minutes.
 	const checkUpdate = async () => ReactDOM.render(await Update(currentVersion), document.getElementById('update-status-container'))
-	setInterval(async () => {
-		await checkUpdate()
+	setInterval(() => {
+		checkUpdate()
```
lukechampine: `setInterval(checkUpdate, 18000000)` :wink:

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Sia-UI",
-  "version": "v1.0.2",
+  "version": "v1.0.0",
```
lukechampine: looks like that slipped in by mistake

droghio: Aren't we working on v1.0.2 of the ui? Is there any reason we are keeping the old version string, it could make user debugging more difficult?

lukechampine: yes, I meant v1.0.2 is the correct version

avahowell: I had manually downgraded the string to verify that the update component was showing given an outdated UI version string and accidentally committed the downgraded string

DavidVorick:
will you get an update popup every 30 minutes? It should only pop up like once a week if you haven't closed the UI. (check every 30 minutes, but skip the notification if it's already been shown to the user).

Maybe instead of once a week it's just one time total.

DavidVorick:
We've decided to close this for now and consider a different approach for v1.1

PR #382, opened 2016-07-27
Terminal Plugin - Add support for loading 033x wallets and siag files

### LINKED ISSUE #380: Terminal Plugin - Load siag + Load 033x Wallet
Either explicitly disable these commands or add their support into the terminal plugin.

droghio:
This update routes the calls `wallet load 033x` and `wallet load siag` to the relevant API endpoints to bypass issues we were having with shell-less password entry on Windows in the same method used for the `wallet unlock` and `wallet load seed` commands. This resolves #380.

Right now paths are passed as is to the API endpoint which could lead to confusion due to https://github.com/NebulousLabs/Sia/issues/1385.

Any thoughts on whether it would be a good idea to disallow paths with `~` or relative paths for the Terminal Plugin?

### DIFF
--- a/plugins/Terminal/js/components/walletpasswordprompt.js
+++ b/plugins/Terminal/js/components/walletpasswordprompt.js
@@ -1,10 +1,9 @@
 import React from 'react'
 import querystring from 'querystring'
-import { httpCommand, commandType } from '../utils/helpers.js'
+import { httpCommand, commandType, getArgumentString } from '../utils/helpers.js'
 import * as constants from '../constants/helper.js'
 
 //This command needs a second prompt.
-const moreSpecialCommands = [ ['wallet', 'load', 'seed'] ]
 export default class WalletPasswordPrompt extends React.Component {
 	componentDidUpdate() {
 		//Give DOM time to register the update.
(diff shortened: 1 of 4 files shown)


plugins/Terminal/js/components/walletpasswordprompt.js
```diff
@@ -17,14 +16,32 @@ export default class WalletPasswordPrompt extends React.Component {
...
 				case constants.WALLET_SEED:
 					this.props.actions.showSeedPrompt()
 					break
 
+				case constants.WALLET_033X:
+					let options = {
+						'source': getArgumentString(this.props.currentCommand, constants.specialCommands[constants.WALLET_033X]),
+						'encryptionpassword': this.props.walletPassword,
+					}
+				case constants.WALLET_SIAG:
```
lukechampine, line 31: please document the fallthrough here. It's sneaky

plugins/Terminal/js/utils/helpers.js
```diff
@@ -156,16 +186,15 @@ export const httpCommand = function(commandStr, actions, newid) {
...
-
-	case 'wallet load seed':
+	} else if (commandString === 'wallet load seed') {
 		apiURL = '/wallet/seed'
-		break
-
-	default:
+	} else if (commandString.indexOf('wallet load 033x') === 0) {
+		apiURL = '/wallet/033x'
+	} else if (commandString.indexOf('wallet load siag') === 0) {
```
lukechampine: doesn't string have an `includes()` method?

droghio: Ah, yeah new in ES6. Old habits die hard I suppose.

plugins/Terminal/js/components/walletpasswordprompt.js
droghio, line 31: Roger, added in latest commit.

plugins/Terminal/js/utils/helpers.js
```diff
@@ -64,6 +64,36 @@ export const commandType = function(commandString, specialArray) {
...
+		args.shift()
+	}
+
+	//Remove command from sting.
+	for (const token of rawCommandSplit) {
+		if (args[0] === token) {
+			args.shift()
+		} else {
+			console.log(`ERROR: getArgumentString failed, string: ${commandString}, did not contain command: ${rawCommand}`)
+			return ''
```
lukechampine: `rawCommand` is not a variable

plugins/Terminal/js/utils/helpers.js
```diff
@@ -64,6 +64,36 @@ export const commandType = function(commandString, specialArray) {
...
+	if (args[0] === './siac' || args[0] === 'siac') {
+		args.shift()
+	}
+
+	//Remove command from sting.
+	for (const token of rawCommandSplit) {
+		if (args[0] === token) {
+			args.shift()
+		} else {
+			console.log(`ERROR: getArgumentString failed, string: ${commandString}, did not contain command: ${rawCommand}`)
```
lukechampine: `rawCommand` is not a variable

PR #383, opened 2016-07-29
Update to sia.js 0.3.0

avahowell:


js/rendererjs/pluginapi.js
```diff
@@ -19,40 +21,43 @@ window.onload = () => {
...
-// Only call siad if siad is running.
-const callWrapper = (uri, callback) => {
-	Siad.ifRunning(() => {
-		Siad.call(uri, callback)
-	})
-}
 
 window.SiaAPI = {
-	call: callWrapper,
+	call: async function(url, callback) {
```
avahowell, line 50: Not really sure about this function.  Since it's an `async` function, we could return the results instead of calling back, but this breaks compatibility with all plugins using SiaAPI.call.  Maybe it's worth it?

js/rendererjs/loadingScreen.js
```diff
@@ -113,13 +85,27 @@ export default function loadingScreen(initUI) {
...
+		}
+
+		// Wait for this process to become reachable before starting the UI.
+		const sleep = (ms = 0) => new Promise((r) => setTimeout(r, ms))
+		while (await Siad.isRunning(siadConfig.address) === false) {
+			await sleep(500)
+		}
+		startUI('Welcome to Sia', initUI)
+	} else {
+		startUI('Welcome back', initUI)
```
lukechampine: in Go style, avoid unnecessary nesting. I would rewrite this as:

``` js
if (running) {
    startUI('Welcome back', initUI)
    return
}
// !running code
```

js/rendererjs/pluginapi.js
```diff
@@ -1,16 +1,18 @@
...
 const mainWindow = remote.getCurrentWindow()
 const config = remote.getGlobal('config')
-Siad.configure(config.siad)
+const siadConfig = config.siad
 let disabled = false
 
-window.onload = () => {
+const sleep = (ms = 0) => new Promise((r) => setTimeout(r, ms))
+
+window.onload = async function() {
```
lukechampine, line 15: why does the `onload` function need to marked `async`? Do you need to mark the function `async` if it uses `await` in the body?

js/rendererjs/pluginapi.js
lukechampine, line 50: I'm in favor of doing so, but not in this PR

js/rendererjs/pluginapi.js
avahowell, line 15: Correct, any functions which `await` must be marked `async`.

js/rendererjs/loadingScreen.js
avahowell: ACK

PR #386, opened 2016-08-04
Send Siad SIGKILL when the UI is quit

avahowell:
Nodejs' child_processes can outlive their parents on Linux and OSX, leading to instances of siad outliving users' UIs.  This PR adds a `quit` event listener which sends siad a `SIGKILL`, preventing it from outliving the UI.  This signal is handled in siad (currently by the API package) and triggers a clean shutdown.

### DIFF
--- a/js/rendererjs/loadingScreen.js
+++ b/js/rendererjs/loadingScreen.js
@@ -98,6 +98,9 @@ export default async function loadingScreen(initUI) {
 		siadProcess.on('error', (e) => showError('Siad couldnt start: ' + e.toString()))
 		siadProcess.on('close', () => showError('Siad unexpectedly closed.'))
 		siadProcess.on('exit', () => showError('Siad unexpectedly exited.'))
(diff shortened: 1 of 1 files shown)


DavidVorick:
Does it kill `siad` in the event that it did not spawn `siad`? Very annoying for a lot of users who want to run siad in the background all the time but only want the GUI open once in a while.

avahowell:
Sia-UI only kills instances of `siad` that it launches.  If a `siad` is already running when the UI starts it is just connected to and not issued any signals on quit.  See https://github.com/NebulousLabs/Sia-UI/blob/8ec89ef3427c0146ffea291e18826ddc0bf1e549/js/rendererjs/loadingScreen.js#L69

DavidVorick:
Oh. Got confused b/c you just added an `onQuit` line of code, now I see that code is only run if you spawn siad.

Thanks.

PR #389, opened 2016-08-15
do not show `available` files as uploading

avahowell:
This PR sets the progress of uploads to 100% once they have become available.  This is per a suggestion from @DavidVorick, since uploads almost never reach full redundancy.

### DIFF
--- a/plugins/Files/js/sagas/helpers.js
+++ b/plugins/Files/js/sagas/helpers.js
@@ -114,11 +114,10 @@ export const totalUsage = (files) => readableFilesize(files.reduce((sum, file) =
 // return a list of file uploads
 export const parseUploads = (files) => List(files)
 .filter((file) => file.uploadprogress < 100)
-.filter((file) => !file.available)
 .map((upload) => ({
 	siapath: upload.siapath,
 	name: Path.basename(upload.siapath),
-	progress: Math.floor(upload.uploadprogress),
+	progress: file.available ? 100 : Math.floor(upload.uploadprogress),
 	type: 'upload',
 }))
 .sortBy((upload) => upload.name)
--- a/plugins/Files/js/sagas/helpers.js
+++ b/plugins/Files/js/sagas/helpers.js
@@ -114,6 +114,7 @@ export const totalUsage = (files) => readableFilesize(files.reduce((sum, file) =
 // return a list of file uploads
 export const parseUploads = (files) => List(files)
 .filter((file) => file.uploadprogress < 100)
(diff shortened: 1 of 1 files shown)


lukechampine:
Hmm. Feels a little underhanded. I'm fine merging this for now, but we should think about a better way to display both values (percentage uploaded and availability). e.g. instead of a percentage, you display "uploading" or "available" and stick the percentage in a tooltip.

DavidVorick:
We had decided not to merge this, I guess we should have made the conversation public. The problem:

```
Loading: | ==>                                           |
         | ===>                                          |
         | ====>                                         |
*poof*
```

The user is confused, because they go from like 35% complete to disappeared loading bar with no indication of completion.

I'm okay with disguising the difference between 'available' and 'fully redundant' in the uploading status. The whole point of the GUI is that we make things as simple as possible, which means hiding some details behind the curtains.

avahowell:
Not quite accurate, as of the latest commit the upload will simply show 100% if it is `available`, and is not removed from the upload list.

DavidVorick:
As discussed offline, I think what would be most intuitive to a user is if it accellerated from wherever it was to 100% over the course of like 3-5 seconds. We can't tell at what percentage a file will become available (technically could be as high as 98%), so we can't cheat the progress too much

Though if we wanted to be really clever, we would have 0-90% represent up to 40% upload, and then 90%-100% represent the other 60%, because most files will become available by 40%. Then it'll just take a long time to get that final bit for larger files or files that aren't uploading well.

lukechampine:
> I think what would be most intuitive to a user is if it accellerated from wherever it was to 100% over the course of like 3-5 seconds.

But that's outright lying, and it will be surprising to see a file slowly upload to 40% over the course of an hour and then jump to 100% in 3 seconds.

I'm fine with displaying "percentage of availability" (though it would be nice to have the true percentage somewhere as well). We would want to add backend support for this instead of trying to fake it, because you really need to know what percentage of chunks have at least 1 host.

DavidVorick:
Is the number of chunks that have hit availability linear though? The user will still see a really unstable progress bar. I guess the progress bar will be pretty unstable anyway.

lukechampine:
it is currently, since the upload algorithm processes chunks in order and tries to upload all of their pieces. But a different algorithm might result in 0% progress for a while and then really fast progress thereafter. IIRC that's why we moved `uploadProgress` to the backend in the first place (instead of just providing a `bytesUploaded` field).

PR #390, opened 2016-08-16
Fix loading bugs

avahowell:
this PR fixes two loading bugs introduced after the upgrade to sia.js `0.3.0`: not having a default `Siad.address` in `config.js` would cause the UI to check for a running instance of Sia at the address `undefined`.  In addition, `loadingScreen`'s `checkSiaPath` function would `reject(false)` instead of `resolve(false)`, causing an unhandled promise rejection and not returning the correct `false` `running` boolean.


lukechampine:
looks like some lint failures

PR #391, opened 2016-08-19
Rename file button

lolsteve:
Adds a pencil icon in the file browser which opens a dialog to rename that file.


DavidVorick:
I haven't looked at it yet but I wanted to say thanks for the pull request!

plugins/Files/js/components/renamedialog.js
```diff
@@ -0,0 +1,34 @@
+import React, { PropTypes } from 'react'
+
+const RenameDialog = ({siapath, actions}) => {
+	const onYesClick = (e) => {
+		e.preventDefault()
+		actions.renameFile(siapath, e.target.newsiapath.value)
+		actions.hideRenameDialog()
```
avahowell: Note that this will cause the rename dialog to hide immediately after the user clicks yes to rename the file.  The alternative is to place `actions.hideRenameDialog` inside the `renameFile` saga, such that the dialog is only hidden when a file is successfully renamed.  I think I'd rather have it be in the saga, but the rest of the team's input on this would be helpful.

plugins/Files/js/reducers/renamedialog.js
```diff
@@ -0,0 +1,16 @@
+import { Map } from 'immutable'
+import * as constants from '../constants/files.js'
+
+const initialState = Map({
+	siapath: '',
+	newsiapath: '',
```
avahowell: `newsiapath` is not used

plugins/Files/js/components/renamedialog.js
```diff
@@ -0,0 +1,34 @@
...
+	const onNoClick = () => actions.hideRenameDialog()
+	return (
+		<div className="modal">
+			<div className="rename-dialog">
+				<div className="rename-text">
+					Enter a new name for {siapath}:
+				</div>
+				<form className="rename-form" onSubmit={onYesClick}>
+					<div className="rename-field">
+						<input type="text" name="newsiapath" required autoFocus defaultValue={siapath} />
```
avahowell: So this rename dialog automatically fills with the current full siapath of the file you're renaming.  This could be confusing if the users select all, then type a different name, and the file moves out of the folder it was in.  I think `RenameDialog` should only rename the _base name_ of the file, you can use `Path.basename` and `Path.dirname` to accomplish this, in the `onYesClick` function.

lolsteve:
Thanks for the feedback. Made changes per your suggestions.

plugins/Files/js/components/renamedialog.js
```diff
@@ -0,0 +1,34 @@
+import React, { PropTypes } from 'react'
+import Path from 'path'
+
+const RenameDialog = ({siapath, actions}) => {
+	const onYesClick = (e) => {
+		e.preventDefault()
+		actions.renameFile(siapath, Path.join(Path.dirname(siapath), e.target.newsiapath.value))
```
avahowell: since the form value is now the basename of the file, not the full path, I think it makes sense to change the input's name to `newname` or something similar.  We should also keep in mind that the semantics of `Path.join` mean that users can do directory transversal operations to their file in the rename dialog, ie renaming '/foo/bar' to '../bar' results in the path '/foo'.  Neat.

avahowell:
LGTM :+1:

PR #393, opened 2016-09-08
call /daemon/stop instead of issuing a sigkill

avahowell:
This PR changes Sia-UI to call /daemon/stop when it is quit, instead of issuing a SIGKILL.  Note that this will trigger clean shutdown of Siad, which currently can take an unreasonably long time.


lukechampine:
could we give siad a few seconds to cleanly shutdown, and then issue SIGKILL?

avahowell:
Yes, but we should probably also add a 'Quitting..' state to the loadingScreen, so the user knows what's going on.

DavidVorick:
How many seconds is a few seconds? 15? I want `siad` to have at least 5 seconds to orchestrate clean shutdown, preferably more like 10 or 15.

Definitely shouldn't merge this until there's some way that the user has obvious indication that shutdown is happening.

avahowell:
agreed, I'll push some commits that add the 'quitting' screen once I get a chance.  Do not merge until that is complete.  I think anything less than 30 seconds is acceptable.

avahowell:
As of 749766b, the UI is now doing clean shutdown.  If the user quits the UI, all plugins are unloaded and a 'Quitting...' overlay is displayed, allowing up to 15 seconds for siad to quit cleanly.  If this timeout elapses, a kill signal is issued to the siad process.  This behavior only applies if the siad process was created by the UI.

DavidVorick:
Is there a way that you can set up automated testing around this?

avahowell:
Yes, I'm setting it up now.

avahowell:
as of 5a532e8, automated testing for the UI's process management of siad is in place.

test/app.js
```diff
@@ -0,0 +1,147 @@
...
+			if (child.COMMAND === 'siad') {
+				process.kill(child.PID, 'SIGKILL')
+			}
+		})
+		resolve()
+	})
+})
+
+// isProcessRunning leverages the semantics of `process.kill` to return true if
+// the input pid is a running process.
```
lukechampine: could use some more explanation of this magic

.travis.yml
```diff
@@ -12,5 +12,15 @@ branches:
...
 
+before_script:
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
+  - sleep 3
+
 install:
+  - wget https://github.com/NebulousLabs/Sia/releases/download/v1.0.1/Sia-v1.0.1-linux-amd64.zip
+  - unzip Sia-v1.0.1-linux-amd64.zip
+  - mv Sia-v1.0.1-linux-amd64 Sia
```
lukechampine, line 23: is there an `unzip` flag for extracting into a directory? Might save you a command. nbd though

avahowell: Not that I could find.

Issue #402, opened 2016-10-10
Old file transfers do not get removed from the File Transfers list


Fornax96:
The file transfers list shows uploads and downloads that are happening. But when an upload or a download finishes the progress bar does not get removed from the list, instead it lingers around. Since the list is ordered by progress all the completed transfers stay at the top of the list, so you have to scroll down to view the progress of the active transfers.

Uploads should either be removed once they reach a redundancy value of 6.00 or 1.00 (I'd day 6.00, since that counts as 100% uploaded in siad). And downloads should simply be removed once the download is complete.

I've noticed that restarting the UI does remove the completed download progress bars. But it does _not_ remove the completed uploads, they will just stick around forever. As a result the upload queue has just become a complete list of every file in the renter.

EDIT: Tested on version 1.0.3-RC

Fornax96:
I've checked the API and all my uploads do seem to have 100% upload progress. I know there used to be a rounding error there in older versions of sia but it has since been fixed. I don't know what caused the uploads to linger around in my queue for so long, but when I arrived at my PC this morning the upload queue was empty. I left the client running overnight and it had finished all my uploads.

avahowell:
Closing this for now.

PR #403, opened 2016-10-10
remove uploads from upload list once they reach >99% progress

avahowell:
Due to some rounding problems, some uploads never reach exactly 100%. This causes them to stay in the upload list unexpectedly, as observed in #402. This PR changes `parseUploads` to filter out files that have more than 99% progress, instead of the previous 100% progress.

### DIFF
--- a/plugins/Files/js/sagas/helpers.js
+++ b/plugins/Files/js/sagas/helpers.js
@@ -113,7 +113,7 @@ export const totalUsage = (files) => readableFilesize(files.reduce((sum, file) =
 // Parse a list of files from `/renter/files`
 // return a list of file uploads
 export const parseUploads = (files) => List(files)
-.filter((file) => file.uploadprogress < 100)
+.filter((file) => file.uploadprogress < 99)
 .map((upload) => ({
 	siapath: upload.siapath,
 	name: Path.basename(upload.siapath),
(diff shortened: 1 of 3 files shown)


lukechampine:
Is the rounding error in the frontend or the backend?

Also, remember to add `fixes #402` or `closes #402` in your commit message so that the issue is automatically closed upon merge

avahowell:
I was under the impression that the rounding error was in the backend, but it looks like they are reaching 100% uploaded now (see Fornax's update to #402). As such, I'm going to close this PR for now.

DavidVorick:
Hmm. I wonder if we should remove the uploads after they hit full availability.

Probably not, as we don't want the user deleting the source file if they haven't achieved full redundancy yet.

Fornax96:
But then the progress bar should also indicate full availability instead of 1.00 redundancy. Currently the progress bar is full when the file is available for download (1.00 redundancy), but it stays in the upload queue for long after that (6.00 redundancy), which makes it look like it's stuck or something.

DavidVorick:
It should be red/yellow when not available (red for <75% to available, yellow for >75% to available), and green when available. But I agree the % progress should be related to % towards full redundancy, and not % towards availability.

PR #406, opened 2016-10-10
improve wallet CSS

avahowell:
This PR improves the wallet plugin's styling by monospacing the transaction id and value in the transaction list, making transaction IDs and values line up visually.

### DIFF
--- a/plugins/Wallet/js/components/transactionlist.js
+++ b/plugins/Wallet/js/components/transactionlist.js
@@ -25,7 +25,7 @@ const TransactionList = ({transactions}) => {
 		}
 		return (
 			<tr key={key}>
-				<td className="txval">{valueData}</td>
+				<td>{valueData}</td>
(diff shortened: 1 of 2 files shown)


plugins/Wallet/js/components/transactionlist.js
```diff
@@ -25,8 +25,8 @@ const TransactionList = ({transactions}) => {
 		}
 		return (
 			<tr key={key}>
-				<td>{valueData}</td>
-				<td>{txn.transactionid}</td>
+				<td className="txval">{valueData}</td>
```
lukechampine: I don't see a corresponding `txval` in the css

avahowell: ah, I originally was formatting the value as monospace as well. I think it only makes sense to monospace the IDs, so I've removed the `txval` class.

PR #407, opened 2016-10-10
add timestamp to transaction list

avahowell:
This PR adds a timestamp to the wallet transaction list.

Screenshot:

txlist

### DIFF
--- a/plugins/Wallet/js/components/transactionlist.js
+++ b/plugins/Wallet/js/components/transactionlist.js
@@ -25,6 +25,7 @@ const TransactionList = ({transactions}) => {
 		}
 		return (
 			<tr key={key}>
+				<td>{txn.confirmationtimestamp.toString()}</td>
 				<td>{valueData}</td>
 				<td className="txid">{txn.transactionid}</td>
 				<td>{txn.confirmed ? <i className="fa fa-check-square confirmed-icon"> Confirmed </i> : <i className="fa fa-clock-o unconfirmed-icon"> Unconfirmed </i>}</td>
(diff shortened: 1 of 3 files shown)


pmknutsen:
To preserve real-estate I would format the timestamp as yyyy-mm-dd hh:mm.
Local timezone should be implied i.e. no GMT-xxxx (EDT).
Rename Timestamp to Date.
Can width of Confirmation Status be reduced? e.g. no column header, just an icon (red or green) to indicate unconfirmed/confirmed?
Rename Net Value to Amount?

lukechampine:
In a lot of file browsers you'll see shortened dates when the date is the same day, i.e. `Today at 4:08 PM` instead of `Tuesday, Oct 11 2016 at 4:08 PM`. Would be a nice touch

avahowell:
latest commit has the following behavior:

if the transaction occured in the last 24 hours, display `Today at HH:MM` timestamp. Otherwise, display `YYYY-MM-DD HH:MM`.

DavidVorick:
screenshot?

avahowell:
timestamp

pmknutsen:
For consistency, can you change "Confirmed/Unconfirmed" to sans font?

pmknutsen:
I still find Net Value confusing. Its ambiguous and could imply Net Sum of the wallet. Amount would be preferred. While on the topic, would it be possible add also a Balance (or, Net Sum) column?

DavidVorick:
I feel like adding a Balance field would be too much information for what we're aiming at.

DavidVorick:
'Amount' probably makes sense.

Eventually, we should have it where you can click on a transaction and get the breakdown of what's happening.

I thought about it more, and maybe reporting the balance as-of that transaction wouldn't be too much clutter, but that would require knowing all of the other transactions, because the API does not have that value I believe.

Issue #333, opened 2016-06-29
Buying 10gb of storage deducted significantly more SC than the quoted price.


AdamBLevine:
I was quoted about 8000sc for 10gb of storage but looking at my balance now, slightly more than 14,000sc was deducted.   If there's a valid reason why this happened it should be explained to me before I click the button that executes the irreversible action.   

There was mention that anything unused would be returned but I have no idea what I've even really bought besides "10gb" - Was that payment for a month?  It should say exactly what the user is buying.

This all falls under expectation management, you need to tell users what's going to happen before they make the decision to do it and then it needs to happen as you've specified and they agreed to.

TacoByte:
#328 and #330

avahowell:
fixed in #409

Issue #330, opened 2016-06-29
Available GB is not what is expected


TacoByte:
I bought 10 GB. I've been playing around uploading a few small files.

The UI is reporting that I have 2GB left!

**582 KB Used / 2 GB Available**

```
Renter info:
    Storage Spending:  606.6 mS
    Upload Spending:   21.02 SC
    Download Spending: 0 H
    Unspent Funds:     4.146 KS
    Total Allocated:   4.167 KS

Tracking 6 files:
 252.0 KB  1453756944023.png
   6.1 KB  untitled folder 3/.DS_Store
   7.8 KB  untitled folder 3/1453756843707.jpg
 252.0 KB  untitled folder 3/1453756944023.png (uploading, 8.33%)
   8.0 KB  untitled folder 3/untitled folder/1443814339804.jpg (uploading, 8.33%)
  56.4 KB  untitled folder 3/untitled folder/1452160563683.jpg (uploading, 8.33%)
```

avahowell:
What is the output of `siac renter allowance`?

TacoByte:
```
Allowance:
    Amount: 1.548 KS
    Period: 12960 blocks
```

avahowell:
Okay, so I've verified that the storage estimation calculation is isomorphic (there's no error in the calculation of how much "storage" you have with a given allowance.)  The reason this happens is that there can be price volatility in the host db.  The average price may be lower or higher when you "buy storage" than when you use the Files plugin.  "Buying storage" is really a leaky abstraction, what you're actually doing is setting a maximum allowance for the UI to spend on storage on your behalf.  This allowance is in coins and can correspond to more or less storage depending on the average hostdb price.

It may be clearer to simply show disk usage and spending instead of trying to calculate storage capacity from the allowance.

TacoByte:
Gotcha. Yeah, I guess this is one more problem that is made worse because of a tiny host ecosystem and high SC price volatility. It might be a good idea to change the whole "buy" idea into something else like what you say. My expectation was definitely that I would effectively buy/reserve/pre-pay a given amount of GB and that I would use it to my liking (and then get refunded what I didn't use).

TacoByte:
I'm now getting `582 KB Used / 2 MB Available`!

avahowell:
fixed in #409

Issue #328, opened 2016-06-29
Inconsistent cost predictions vs transactions vs renter info


TacoByte:
Darwin 1.0

I bought 10GB of storage. It was quoted at around 1500 SC.

My wallet then sent many transactions, with a total well in excess of the original quote, around 10000 SC.  This [screenshot](http://i.imgur.com/ouorPvo.jpg) shows most of the transactions that were sent in this batch.

This is what renter returns, which seems inconsistent as well (the 5346 transaction alone seems larger than the allocated funds)

```
Renter info:
    Storage Spending:  25.43 mS
    Upload Spending:   20.97 SC
    Download Spending: 0 H
    Unspent Funds:     3.932 KS
    Total Allocated:   3.953 KS

Tracking 1 files:
 252.0 KB  1453756944023.png (uploading, 8.33%)
```

I don't know if this is "just" the UI misreporting things, or if something more lower level is going on.

DavidVorick:
The renter is saying that it only used 4,000 coins between all of the contracts. Not sure why that's such a gap from 1500.

The wallet is reported that 10,000 coins have been spent?

TacoByte:
If I add up all the transactions that happened when I bought the 100GB, I get around 7000 SC of spending (not quite 10000, sorry for rounding up so aggressively :) but still far away from 1500)

DavidVorick:
Ok. That's definitely not right, I have some ideas about why though.

DavidVorick:
The contract formation right now is failing with a large percentage of the hosts, which means that the price estimation is off. The price estimation and formation use the cheapest hosts, but if too many cheap hosts fail, you will switch to the more expensive hosts, resulting in a bad estimation.

TacoByte:
I was actually suspecting something like that. The host market is so illiquid right now that it results in a lot of granularity and discrepancies between what Sia predicts will happen based on available information and what deals actually do get passed on the market/blockchain.

TacoByte:
That still doesn't explain why renter reports spending 4 KS when my outgoing transactions add up to more than that?

Also, I uploaded some more files (well within my already concluded 10GB contract), and more SC is being sent from the wallet? -584.58 I was expecting to not have to spend any more once I had signed the contracts?

superdcc:
Same Problem
10 GB, Estimated 1300,But really cost 7500 
And still show : not enough contracts to upload files  (when upload file)

[ScreenShot](http://imgur.com/JS3Vx4A)

[...]

PR #409, opened 2016-10-12
Renter 'Buy Storage' UX improvements

avahowell:
This PR addresses the UX problems mentioned in issues #333, #330, and #328. Since host prices are volatile and subject to change, presenting the user with 'storage plans' by estimating the storage available from a certain allowance leads to all of these unexpected behaviors. Estimating the user's remaining storage is subject to the same problem. This has been addressed by removing the leaky 'Buy Storage' and 'Storage Available' abstractions; now the user is directly asked for an allowance to set and usage is reported as 'x SC spent / x SC allocated'.

In addition, a 'contractor status' component has been added, which displays the number of contracts the renter currently has, or a 'Forming Contracts...' loading spinner if the renter is currently setting the user's allowance.

Screenshots:
sia1
sia2
sia3
sia6
sia8


VoidingWarranties:
We spent so much time working on the leaky abstraction and now we're getting rid of it 😭

It seems like the leaky abstraction is sinking like the Titanic. Overall I think this is a great change. It might be jarring for new users, but it's not like SC spent / SC allocated is too difficult to understand.

My only concern is that now a user has no sense of how many SC they should allocate when they buy storage. Perhaps we should show them a little helper text saying something like "The market price for 10GB is x SC right now (but the market is volatile and blah blah blah)"

avahowell:
Latest commit adds storage estimation as a footnote to the 'buy storage' screen:

sia-estimate

lukechampine:
Agreed, the purpose of the leaky abstraction was to frame things in terms of storage rather than money.
Actually, showing only the storage cost may be misleading, since download cost usually exceeds storage cost. It's just a hard thing to estimate unless you know the user's usage patterns. :|

avahowell:
Note that the 'Spent' cost which is shown is the sum of upload, download, and storage spending.

pmknutsen:
I find the wording in the input dialog a bit confusing, where it says "... After 12 weeks the contractor will renew the contracts it forms.". This seems to imply that funds will be automatically withdrawn from the wallet after 12 weeks, but I was under the impression that renewal of storage contracts is something the user always has to initiate (unlike files, which are automatically renewed)?

Also, I would avoid terms such as "contractor" as few non-power users will have any idea what that is. It would be sufficient to say "Sia".

DavidVorick:
Also needs to mention fees.

"Before you can upload to Sia, you have to form contracts with the hosts. This creates a pool of money you can spend on uploading and downloading. Funds stay in the pool for 3 months. Funds not spent after 3 months will be returned to your wallet. Fees will not be returned.

The pool will refill automatically every 6 weeks, using the full 3 month budget. You will pay the fees again. Your computer must be online to refill the allowance, and if the allowance is unable to refill before the pool expires, you will lose all of your data."

Set Storage Allowance: [ 5000 SC]
Estimated Fees: 1600 (1000 + 0.12 \* allowance)

Estimated Prices:

Upload: 100 SC / GB
Download: 2500 SC / GB
Storage: 800 SC / GB / Mo

(or whatever the prices are)

DavidVorick:
Should there be a progress bar for the 'spent' vs. 'allocated' part?

avahowell:
Perhaps. If we did that, I'd want to have a tooltip that shows the numbers as well.

Screenshot of the latest iteration:
newbuystorage

pmknutsen:
@DavidVorick I found the text a bit too technical (and unclear):

"You need to allocate funds for uploading to Sia. Your storage allowance remains locked for 3 months. Unspent funds are then refunded*. You can add funds to your allowance later.

Your storage allowance automatically refills every 6 weeks, up to your total storage allowance. Your computer must remain online, and the wallet unlocked, to complete automatic refills. If Sia fails to refill the allowance by the end lock-in period, your storage contracts will not be renewed and you risk loosing some or all of your data.
- transaction fees are non-refundable."

@johnathanhowell Can you increase the margins of the dialog?

DavidVorick:
Also, dialogs with that much text should be left-aligned. The paragraph should be centered with padding but within the paragraph the text should be left-aligned.

DavidVorick:
I like the text by @pmknutsen more than my own text, but still seems like it could use some work. @lukechampine perhaps you want to take a stab?

lukechampine:
I like @pmknutsen's copy too, although I would make it more clear that the funds are used for bandwidth as well as storage costs (also: loosing -> losing). Perhaps some of the minor details (like the transaction fees bit) could be moved to a tooltip to prevent overwhelming the user with information.

DavidVorick:
"You need to allocate funds to upload and download on Sia. Your allowance remains locked for 3 months. Unspent funds are then refunded*. You can increase your allowance at any time.

Your storage allowance automatically refills every 6 weeks. Your computer must be online with your wallet unlocked to complete the refill. If Sia fails to refill the allowance by the end lock-in period, your data may be lost.

*contract fees are non-refundable."

definitions note for support team - "contract fees" -> transaction fee (scales with transaction size) + contract fee (fee paid per host for forming contracts) + siafund fees (0.39% of storage + collateral)

pmknutsen:
"...by the end lock-in period" should be "...by the end of the lock-in period"

avahowell:
updated with the new copy, along with some minor styling changes. Here's the screenshot:

sia

avahowell:
Should the User be prevented from setting an allowance if an allowance is currently in the process of being set (ie: the UI is displaying "Forming contracts...")? I was thinking it would probably be advisable to disable the 'Buy Storage' button while contracts are being formed. Also, does 'Buy Storage' still make sense as a name for the button, now that the user is directly setting their allowance?

DavidVorick:
Margins should be maybe 2x as large, and the asterisk text should be slightly smaller, and it should be left-aligned with the other paragraphs. The spacing between the paragraphs should probably be reduced.

I think it makes sense to prevent the user from setting an allowance while contracts are being formed. Might annoy them because it can take up to 30 minutes to form contracts, but it'll prevent some potential headaches for the renter, which may not safely handle such an event.

Probably 'Set Allowance' or 'Create Allowance' or 'Update Allowance'

The Estimated Fees / Estimated Storage based on current prices should be tabular, such that the numbers align. Within the cells, text should be left-aligned and values should be right-aligned. I think I'm in favor of displaying the rows "Fees / Cost of Storage / Cost to Download", and then factor the cost of uploading into the cost of storage.

I also think the current visual for the set-allowance field could use work. I'm not sure how to fix it but it currently feels cluttered.

avahowell:
Updated with most recent review comments:

sia

pmknutsen:
Add some space between the input field and SC.

Also, its not clear if Estimated Fees are included in the allowance set by the user, or if they come in addition.

avahowell:
Added spacing, clarified where the fees are paid, and changed the estimation table to be horizontal as @DavidVorick suggested offline:

sia

PR #410, opened 2016-10-13
Multi-file selection, deletion, downloading

avahowell:
This PR adds multi-selection features to the file manager. Files can be selected by clicking on them, and multiple files can be selected by clicking and holding Shift. Files can be deselected by pressing escape.

Screenshots:
sia0
sia1
sia2


VoidingWarranties:
Whoa. :+1:

avahowell:
note that batched downloads do not currently work because of the download exclusion bug that @lukechampine is working on (downloads will sometimes fail with `already revising contract`).

DavidVorick:
The downloads don't go into a queue?

DavidVorick:
(downloads don't go into a queue, but backend updates to support that are in progress)

Need to be able to shift-unclick to deselect, so if you have multiple selected and you select another one on accident, you can deselect one instead of having to deselect all.

avahowell:
shift-clicking selected files now deselects them.

PR #427, opened 2016-10-19
Revised upload progress behavior

avahowell:


DavidVorick:
I think the whole flow between the file browser and the file transfer bar needs to be reconsidered.

For example, it should indicate to the user that their upload is available. We talked about clarifying that with colors, not sure how to pull it all together though.

Also, at what point does the file get added to the file browser? Should the file be in the 'transferring...' bar even if it's mostly complete / available? Those final percents can take a long time, especially because sometimes hosts will have disk failures, and then the file is basically guaranteed not to hit 100%. We can't have files perpetually in the 'transferring' bar simply because they are stuck at 87.5% (which is where my files are currently stuck at).

Redundancy is probably something that should be reflected in the file browser itself, because its a continuum, and not really a single action that can complete.

avahowell:
Files appear in the file list once they reach `available`. Before that, they only appear in transfers. Currently, the upload will unexpectedly show 100% progress once available, leading users to believe that they are stuck in the transfers list (we introduced this behavior here: https://github.com/NebulousLabs/Sia-UI/pull/389). I think the UI should communicate as clearly and accurately as possible; having the progress bars reflect the exact progress of the uploads seems the most correct to me. 

The existence of a file in the file list indicates its availability, the transfers list is only concerned with conveying the progress of uploads and downloads.

VoidingWarranties:
@johnathanhowell I like this change because it is more truthful. A user expects to be able to delete their local file as soon as the progress hits 100%.

@DavidVorick to expand on the above, not only would a user expect to be able to delete their local file as soon as the progress hits 100%, they would also expect to be able to delete their local file as soon as it's removed from the transfer list. If files are building up in the transfer list because of slow progress in the final percentages we should use that as an incentive to improve the upload speed of the final percentages :)

DavidVorick:
@VoidingWarranties sometimes the final percentages are slow because you've got a small number of really slow hosts in your host list. Then the slowdown is entirely unavoidable, because you still need to get the full set of pieces to that host. We could potentially get around it by forming 30 contracts instead of 24, which could also have other advantages, but I'm not sure how big of a change that would be.

VoidingWarranties:
Is it possible to form new contracts with other hosts and abandon a contract with a slow host? If the abandoned contract's money is refunded at the end of the contract, the only downside is the lost time spent uploading and funds temporarily locked in the contract.

We could also consider temporarily banning the renter from creating new contracts with that host (temporarily in case it's just ephemeral congestion).

VoidingWarranties:
Another solution, albeit rather circuitous, would be to add a "stalled transfers" section. If you could collapse this section you could declutter the transfer list without being disingenuous to the user.

nebulousmediacenter:
It's definitely confusing to me that I would upload a file and have it not immediately appear in the file browser. I think that it should appear in the file browser as well as being added to the transfer list, and then there should be some indication of progress/health on the file in the file browser.

I'm also thinking it makes sense to cleanly separate the uploads from the downloads. There should probably be a transfer list somewhere only showing things that are still in progress for uploading, but I don't think the download list is the right place to put them.

avahowell:
The transfers list separates uploads and downloads; it's composed of two lists, an `UploadList` and a `DownloadList`. The `DownloadList` shows only files downloading and comes before the `UploadList`. They're separately labeled with `Downloads` or `Uploads` headers in the filetransfers list. Either list only appears if you have downloads or uploads.

Perhaps we should remove the upload list altogether, rename file transfers to downloads, and show upload progress in the file browser.

DavidVorick:
New proposal:
- files are listed in the task list when downloading, labled as complete when the download finishes (current behavior, desired)
- files are listed in the task list when uploading, labeled as 'Uploading' if redundancy is below 1x, and labeled as 'Boosting Redundancy' if redundancy is below 2.5x
- files are displayed in the file browser immediately, but are faintly highlighted red if they are not available, and are faintly highlighted yellow if redundancy is below 2.5x. No highlights if redundancy is higher than 2.5x
- Files are not displayed in the task list if redundancy is above 2.5x.

Rationale: users really want to get their min redundancy to at least 2.5x at this point, to avoid any potential file loss if they delete their file before the uploading completes and then some hosts go offline. But after that, it's probably safe to let the download-and-reupload process handle the redundancy management.

It should be noted that at this time, the renter does not actually support download-and-reupload management, so users will have to handle that themselves. We will be adding it very soon though.

avahowell:
Ok, I'm going to push more commits on to this PR to implement the new proposal.

avahowell:
Ok, here's a short summary of the current behavior:
- File browser now shows all files, not just `available` files.
- Next to filesize, there is a `redundancy` info icon that displays the current redundancy, and a helpful icon. The icon is red when the file is not available or has a redundancy <1, and it is increasingly dark shades of our canonical green color as it goes from 1.0 -> max redundancy.
- Uploads and downloads both have a `status` prop. This `status` prop changes from 'Uploading', to 'Boosting Redundancy' once the file reaches >1.0 redundancy. The progress bar mirrors exactly the `uploadprogress` reported by the backend, but the new `status` information communicates clearly the state of the upload. Uploads are removed from the upload list once they reach >2.5 redundancy.

Screenshots (taken from ant-farms with various states):
sia2
sia1
cubes
redundancy

It's important to note that directories do not display the redundancy indicator. We could take an average of the redundancy in all files in the directory, but I chose not to implement that right away.

DavidVorick:
For directories you'd want to display the redundancy of the least redundant file

DavidVorick:
Wrong button

For directories, use the whole tree when determining least redundant

DavidVorick:
4 icons + file size is too much

avahowell:
ok, I've added redundancy for directories. We could simply remove the delete and download icons, since that functionality is covered by the new selection mechanism

lukechampine:
Agreed, download/delete can be removed. Users will be expecting the same interactions they learned from their file browser:
- Click once to select
- Click-and-drag, click-and-shift-click, or click-and-ctrl-click to select multiple
- Double click to open (download in our case)
- Double click a folder to enter it
- "Slow double click" to rename
- With selected file, `Del` to delete
- With selected file, `Enter` to open/download
- Right click displays list of actions

etc.

DavidVorick:
I'm thinking that the rename button can be removed as well, then it's just showing the file size and the redundancy factor.

I'm not convinced that the stack of blocks is the best way to say "redundancy", though I'm not sure what else might work well. @pmknutsen @fornax @chehands @VoidingWarranties wondering if you guys have ideas.

avahowell:
I think that removing the extraneous icons and adding the interaction features mentioned by @lukechampine is different enough to warrant its own PR.

DavidVorick:
Okay, we can probably merge this and then bikeshed over the redundancy icons in another PR.

PR #428, opened 2016-10-19
add timeouts, timeout error handling

avahowell:
This PR coincides with https://github.com/NebulousLabs/Nodejs-Sia/pull/36; it changes all polling API requests to log errors to console instead of popping up error dialogs, preventing timeouts from causing unwanted error messaging.  I also cleaned up the plugin API's `call` wrapper function, and added some sane default timeouts for long running requests (wallet unlocking, downloading, etc).

### DIFF
--- a/plugins/Files/js/sagas/files.js
+++ b/plugins/Files/js/sagas/files.js
@@ -5,7 +5,7 @@ import fs from 'fs'
 import * as actions from '../actions/files.js'
 import * as constants from '../constants/files.js'
 import { List } from 'immutable'
-import { ls, allowancePeriod, allowanceHosts, estimatedStorage, totalSpending, handleError, siadCall, readdirRecursive, parseDownloads, parseUploads } from './helpers.js'
+import { ls, allowancePeriod, allowanceHosts, estimatedStorage, totalSpending, siadCall, readdirRecursive, parseDownloads, parseUploads } from './helpers.js'
 
 
 // Query siad for the state of the wallet.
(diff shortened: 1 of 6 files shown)


DavidVorick:
we are intentionally not merging this until after v1.0.3 is released because the potential surface area for bugs here is pretty large, and we don't have enough time to properly test it.

then we can merge it and get it into the next version.

plugins/Files/js/sagas/files.js
```diff
@@ -132,6 +134,7 @@ function* downloadFileSaga(action) {
 		if (action.file.type === 'file') {
 			yield siadCall({
 				url: '/renter/download/' + encodeURI(action.file.siapath),
+				timeout: 60000,
```
lukechampine: this should be much longer; something ridiculous like 1 week (or more).

Although, I'm not sure how soon we plan to make the download call non-blocking. If we can get that in before the next release, then we wouldn't need a huge timeout here. Still, it's probably better to change it twice than to risk forgetting about it.

avahowell: I added some zeros to this timeout :smile:

plugins/Files/js/sagas/files.js
```diff
@@ -65,6 +65,7 @@ function* setAllowanceSaga(action) {
 		yield siadCall({
 			url: '/renter',
 			method: 'POST',
+			timeout: 18000000, // 30 minute timeout for setting allowance
```
DavidVorick: That's probably not going to be long enough, as the minimum is I think 45 seconds per contract. Add in enough failures and it's going to take more than 30 minutes.

Probably should be 120 minutes.

plugins/Files/js/sagas/files.js
```diff
@@ -134,7 +134,7 @@ function* downloadFileSaga(action) {
 		if (action.file.type === 'file') {
 			yield siadCall({
 				url: '/renter/download/' + encodeURI(action.file.siapath),
-				timeout: 600000000,
+				timeout: 6e8,
```
DavidVorick, line 137: 600e6

PR #430, opened 2016-10-19
use current time as timestamp if transaction is not confirmed

avahowell:
This fixes unconfirmed transactions having `NaN NaN NaN` timestamps


lukechampine:
Can we use "Not confirmed" instead? Or does it rely on the timestamp for proper sorting or something?

avahowell:
We can use not confirmed, it's probably the more correct choice anyway.

Issue #416, opened 2016-10-17
siad dependency not documented


DavidVorick:
If you are installing the UI from source, you need to have siad binary to talk to, as it's not included in the repo. This should be documented in the readme.

Currently it won't check your path for siad. This might be the best thing to do, or maybe it should check in your path.

Issue #415, opened 2016-10-17
libxss.so.1 dependency


DavidVorick:
on Debian, it seems that you need to install libXss.so.1, which can be done via `sudo apt-get install libxss1`.

This should probably be documented in the readme.

avahowell:
It should be noted also that this is a dependency of `electron`.

PR #434, opened 2016-10-21
update readme with siad dependency, libxss, npm/nodejs version

### LINKED ISSUE #413: Specifiy minimum npm/nodejs version in README
The README does not specify which version of node is needed to run Sia-UI, but I seem to recall that it's higher than the latest LTS release (v4.6.0).

avahowell:
Closes #413, #415, #416

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,10 @@ the Sia network.
 
 ## Prerequisites
 
-- [node & npm (packaged together)](https://nodejs.org/download/)
+- [siad](https://github.com/NebulousLabs/Sia)
+- [node & npm 6.9.0 LTS](https://nodejs.org/download/)
+Earlier node versions may work, but they do not have guarenteed support.
+- `libxss` is required for electron on some systems.
 
 ## Running
 
@@ -26,6 +29,7 @@ the Sia network.
 
 Run from source
 
+0. Install dependencies mentioned above
 1. Download or `git clone` the repository
 2. npm install
 3. npm start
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,7 +16,10 @@ the Sia network.
 
 ## Prerequisites
 
-- [node & npm (packaged together)](https://nodejs.org/download/)
+- [siad](https://github.com/NebulousLabs/Sia)
+- [node & npm 6.9.0 LTS](https://nodejs.org/download/)
+Earlier node versions may work, but they do not have guarenteed support.
```
DavidVorick, line 21: Do our builds commit to using the LTS version of node? Want to make sure there will be a failure if something accidentally uses a new feature.

README.md
```diff
@@ -16,7 +16,10 @@ the Sia network.
...
+- `libxss` is required for electron on some systems.
```
DavidVorick, line 23: I would go as far as saying 'On debian, libxss is required as an electron dependency. This can be installed with `sudo apt-get install libxss1`'. The user shouldn't need to figure out what package they need for libxss. Also, better I think to be more specific than 'some systems', as people on uncommon distros usually understand that they could run into similar problems. Our documented use cases should cover all common distros, we will see more issues reported if we are missing things for Ubuntu, fedora, etc.

README.md
avahowell: Our builds currently use node `5.9.0`. I'll update .travis.yml to use `6.9.0`.

Issue #426, opened 2016-10-19
In the hosting window, the 'free' amount interferes with the 'max' amount


Chehands:

DavidVorick:
you might just need to make the window larger, at some point I think we can't support tiny resolutions in the UI.

DavidVorick:
Should be fixed now. Thanks again for creating all of these issues, definitely helps us to make sure everything is resolved without forgetting anything.

PR #439, opened 2016-10-21
round Free and Max, fix some misc issues with Hosting's FileList

### LINKED ISSUE #437: rounding/graphical problem in host plugin
screenshot should explain everything
graphical glitch

avahowell:
Fixes #437 and #426. I also noticed the hosting plugin is largely missing proptypes, I'll be adding a linting rule to cause build errors in the future if proptypes are missing and I'll be going through and adding proptypes.


DavidVorick:
Should I leave this open until the linting rule is added?

avahowell:
No. adding the linting rule will add many linting failures and require adding PropTypes anywhere they are missing. It's a large enough change to warrant its own PR.

PR #442, opened 2016-10-25
Replace 100% progress bar with 'Completed' when downloads finish, add 'clear downloads' button

avahowell:


plugins/Files/js/reducers/files.js
```diff
@@ -84,7 +86,7 @@ export default function filesReducer(state = initialState, action) {
 	case constants.RECEIVE_UPLOADS:
 		return state.set('uploading', action.uploads)
 	case constants.RECEIVE_DOWNLOADS:
-		return state.set('downloading', action.downloads)
+		return state.set('downloading', action.downloads.filter((download) => Date.parse(download.starttime) > state.get('showDownloadsSince')))
```
lukechampine, line 89: hmm, does this belong here? You know better than I do, but I lean towards filtering data as late as possible. I guess you wouldn't want to filter it inside the component though; there's no point.

PR #456, opened 2016-10-27
make upload, rename, and delete file api calls serial

### LINKED ISSUE #455: OSX 256 open file limit
Mac OS X has a limit of 256 open files. Each API call counts as an open file. The timeouts pr #427 will help with this, but we still need to queue API calls for operations like folder upload, or else the calls will start failing.

avahowell:
This PR introduces an `actionChannel` to the Files plugin, making upload, rename, and delete API calls occur serially. This fixes a bug found by wskeen, reported in slack. Ideally download calls would also be serial, but they are currently blocking calls in the API.

fixes #455

### DIFF
--- a/plugins/Files/js/sagas/files.js
+++ b/plugins/Files/js/sagas/files.js
@@ -1,5 +1,5 @@
 import { takeEvery } from 'redux-saga'
-import { put } from 'redux-saga/effects'
+import { put, actionChannel, take } from 'redux-saga/effects'
 import Path from 'path'
 import fs from 'fs'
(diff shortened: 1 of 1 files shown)


lukechampine:
We should revisit this sometime, since siad should be able to support concurrent actions. Unless the problems are entirely UI-side?

DavidVorick:
The problem lies with the max number of allowed open files on Mac. It's something tiny like 256. So if we do a bunch of calls in parallel, Mac users start getting nasty error messages and weird failure modes.

PR #458, opened 2016-11-02
replace back button with directory info bar

avahowell:
This PR adds a new component, `DirectoryInfoBar`, that replaces the back button in the file list. This bar is always the first element of the list. It shows the current path along with the number of files in the current directory, and a back button. The back button is greyed out if the user is in the root directory. This fixes the issue where double-clicking the back button would trigger a doubleclick event on a folder in the parent directory, causing the user to go back and forward in quick succession.

### DIFF
--- a/plugins/Files/css/files.css
+++ b/plugins/Files/css/files.css
@@ -286,7 +286,7 @@ div[tabindex="1"]:focus {
 	align-items: center;
 	justify-content: flex-start;
 }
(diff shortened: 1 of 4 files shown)


DavidVorick:
Screenshots

avahowell:
sia
sia2
sia3

^ screenshots of the new behavior.

DavidVorick:
utACK

PR #460, opened 2016-11-04
add notification badge on download or upload

### LINKED ISSUE #459: When i select a file for download, after the destination is confirmed, there is no dialogue to tell me that my download will commence.

avahowell:
This PR adds an animated notification badge that appears after a upload or download on the File Transfers button, in order to notify the user that a download or upload has been initiated. Closes #459

notify1
notify2

### DIFF
--- a/plugins/Files/css/files.css
+++ b/plugins/Files/css/files.css
@@ -209,14 +209,38 @@ div[tabindex="1"]:focus {
 	width: 100px;
 	cursor: pointer;
 }
-.files-toolbar .buttons div {
-	text-align: center;
-	margin-bottom: 5px;
-	width: 100px;
-	cursor: pointer;
-}
 .transfers-button {
 	display: inline-block;
+}
+
+@keyframes notify-animation {
+	from {
+		right: 50%;
+		top: 50%;
+		width: 100px;
+		height: 100px;
+		font-size: 28px;
+		text-align: center;
+	}
+	to {
+		right: 25px;
+		top: 5px;
+		width: 13px;
+		height: 13px;
+		font-size: 12px;
+	}
+}
+
+.transfers-button > .badge {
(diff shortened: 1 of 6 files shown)


lukechampine:
That red is too harsh.
We need another good color for our color scheme.

pmknutsen:
The Sia green is called Caribbean Green. Its complementary color is Venetian Red (#cc0033) which has a slight magenta hue to it. Try that. See http://www.colorhexa.com/00cc99 for some color schemes.

Edit: Why is the confirmation dialog so large?

DavidVorick:
@VoidingWarranties @Chehands do you guys think that this is an acceptable way to notify the user for downloads?

I worry that it is too subtle, on the other hand perhaps it is just the right amount of subtle.

avahowell:
re: subtlety: it's worth noting that there is a 'bounce' animation on the badge where the badge starts out at 2x its size, then shrinks to its shown size in the screenshot, in order to grab the user's attention.

DavidVorick:
screenshots

Chehands:
Looks good to me, does the notification badge stay there all time the files are uploading and then disappear when they've finished?

avahowell:
Currently, the notification badge is cleared when you click the 'File Transfers' button, but maybe it should just be an indication of the number of files currently being downloaded or uploaded.

avahowell:
As of latest commit:
- File transfers button shows an unread badge when an upload or download is started
- The badge is cleared when the user clicks the file transfer button
- If a download or upload finishes and is removed from the upload/download list before the user clicks the file transfers button, the badge will be decremented to reflect the current unread count.

screenshots:

sia2
sia3

Chehands:
I like the idea of its staying there as an indicator of current transfers. The Venetian red looks good.

Chehands:
I think that works really well. Look forward to the release.

PR #463, opened 2016-11-07
only deduplicate files in ls if they have the same type

avahowell:
this PR fixes a bug in the file browser's `ls` function where if folders and files held the same name, one of them would be deduplicated and removed from the file list. `ls` now checks not only for equivalent filenames, but also equivalent types.


DavidVorick:
Could you provide an example? Is it possible to have a file named 'asdf' and a folder also named 'asdf'? That strikes me as something we should fix.

avahowell:
if you had a file named asdf and a folder named asdf, which one appeared in the file list would be whichever appeared in the un-parsed list first. This PR fixed that and was merged, it is now possible to have a file with the same name as a folder.

PR #464, opened 2016-11-08
improve directory uploading robustness

avahowell:
This PR adds `uploadDirectory`, a helper function to make folder upload action generation more testable and robust. Both win32 and posix upload testing has been added.

### DIFF
--- a/plugins/Files/js/sagas/helpers.js
+++ b/plugins/Files/js/sagas/helpers.js
@@ -130,9 +130,9 @@ export const readdirRecursive = (path, files) => {
 	return filelist
 }
 
-// uploadDirectory takes a `source` and a `siapath` and returns a list of
-// upload actions that will upload every file located under `source`/ to
-// `siapath`/.
+// uploadDirectory takes a `directory`, a list of files inside the directory,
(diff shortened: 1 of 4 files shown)


plugins/Files/js/sagas/helpers.js
```diff
@@ -129,6 +130,19 @@ export const readdirRecursive = (path, files) => {
...
+// upload each file to `destpath/directoryname/`.
+export const uploadDirectory = (directory, files, destpath) =>
+	files.map((file) => {
+		const relativePath = Path.dirname(file.substring(directory.length + 1))
+		const siapath = Path.posix.join(destpath, Path.basename(directory), relativePath)
+		return {
+			source: file,
+			siapath,
+		}
+	}).map((upload) => actions.uploadFile(upload.siapath, upload.source))
```
lukechampine: not sure I understand why there are two `map`s here -- clarity?

avahowell: nah, I wrote this function by refactoring the old `uploadFolderSaga` which had two maps. One is definitely the cleaner choice

Issue #449, opened 2016-10-26
Logs Plugin


DavidVorick:
There should be a plugin that allows the user to view the logs. This includes any UI logs, siad stdout+stderr, and any logs in the sia/ folder. (consensus.log, etc.)

This plugin should be hidden by default, but easily brought up with either a hotkey or through some sort of settings window.

PR #467, opened 2016-11-11
add Logs plugin to view Sia logs in UI

avahowell:
This PR implements #449, and the currently open project on this repo. It adds a `Logs` plugin, which can be used to view each siad log, or combinations of the logs, defined by `filters.js`.

Screenshots:
sia1
sia2
sia3
sia4


js/rendererjs/plugins.js
```diff
@@ -116,35 +116,31 @@ export const scanFolder = (path) => {
...
+	plugins = pushToBottom(plugins, 'About')
 
 	// Push the home plugin to the top
-	plugins = plugins.sort((p1) => {
-		if (getPluginName(p1) === homePlugin) {
-			return -1
-		}
-		return 0
-	})
+	plugins = pushToTop(plugins, homePlugin)
```
lukechampine, line 147: this is an interesting way of doing things, haha.
Maybe instead we could map each plugin to a "priority" number?
The problem there, I suppose, is that you want an easy way to stick plugins at _either end_ of the list, not just the top (or just the bottom). So a priority list isn't a perfect fit because it tends to only have one "maximum" value (e.g. 0 means highest priority).

plugins/Logs/js/components/filtercontrols.js
```diff
@@ -0,0 +1,42 @@
...
+}
+
+const FilterControls = ({logFilters, actions}) => (
+	<div style={filterControlsStyle}>
+		{
+			filters.map((filter, key) => (
+				<FilterControl
+					name={filter.name}
+					filters={filter.filters}
+					checked={filter.filters.reduce((isChecked, filtertext) => isChecked || logFilters.includes(filtertext), false)}
```
lukechampine, line 25: congrats, "filter" is now semantically satiated.

plugins/Logs/js/components/filtercontrol.js
```diff
@@ -0,0 +1,43 @@
...
+		}
+		if (checked) {
+			removeLogFilters(clickedFilters)
+			return
+		}
+		addLogFilters(clickedFilters)
+	}
+	return (
+		<div style={filterControlStyle} onClick={onFilterClick(filters)}>
+			<span style={{WebkitUserSelect: 'none'}}> {name} </span>
```
lukechampine, line 28: what does this style do?

avahowell: It prevents the button names from being selected when you shift-click.

plugins/Logs/js/components/logview.js
```diff
@@ -0,0 +1,28 @@
...
+	overflowY: 'scroll',
+	whiteSpace: 'pre',
+	fontSize: '12px',
+	fontFamily: 'monospace',
+}
+
+const LogView = ({logText}) => (
+	<div style={logViewStyle}>
+		{logText}
+	</div>
```
lukechampine: too complicated

lukechampine:
My main concerns are:
- When does the log output get refreshed?
- Is it dangerous to read the entire log file? What if the file is 10MB?

I think ideally the log files would be treated as streams, rather than reading them all into memory. Also, it would be nice if it were possible to only show the output from the current session (maybe via a toggle). Otherwise if you have a long-running daemon you can wind up reading tons of data that you don't care about. Not a pressing concern though.

DavidVorick:
Can we move the bar to the top?

DavidVorick:
LOGS: m1 | m2 | etc

DavidVorick:
DEBUG logs can hit hundreds of MBs

avahowell:
As of latest commit:
- Only the last `n` bytes of logs are loaded into memory. `n` defaults at 50KB.
- If you hit the top of the log view scrolling, 50 more KB of logs will be loaded.

plugins/Logs/js/components/filtercontrol.js
```diff
@@ -0,0 +1,43 @@
...
+	const onFilterClick = (clickedFilters) => (e) => {
+		if (!e.ctrlKey && !e.shiftKey) {
+			setLogFilters(clickedFilters)
+			return
+		}
+		if (checked) {
+			removeLogFilters(clickedFilters)
+			return
+		}
+		addLogFilters(clickedFilters)
```
lukechampine, line 24: I think making each filter toggle would be more natural. If I see a row of buttons, my UX-sense doesn't expect that I'll be able to shift-click them.

plugins/Logs/js/components/logview.js
```diff
@@ -0,0 +1,51 @@
...
+	whiteSpace: 'pre',
+	fontSize: '12px',
+	fontFamily: 'monospace',
+}
+
+export default class LogView extends React.Component {
+	componentDidMount() {
+		this._logView.scrollTop = this._logView.scrollHeight
+		this._logView.addEventListener('scroll', this.handleScroll.bind(this))
+	}
```
lukechampine, line 21: tragic

plugins/Logs/js/logparse.js
```diff
@@ -0,0 +1,48 @@
+import fs from 'fs'
+import { readdirRecursive } from './utils.js'
+
+const siadir = SiaAPI.config.siad.datadir
+
+const fileSize = (logpath) =>
+	fs.statSync(logpath).size
```
lukechampine: I usually stick short arrow function bodies on the same line if they're just a single expression. It's arguably less readable, but with a newline+indent it looks like python :P

avahowell: yeah, in this case I think a single line is prettier.

plugins/Logs/js/logparse.js
```diff
@@ -0,0 +1,48 @@
...
+
+// readLog takes a log path and returns a string representing the log data.
+export const readLog = (logpath, nbytes) => {
+	const len = nbytes || fileSize(logpath)
+	const startPos = Math.max(0, fileSize(logpath) - len)
+	const buf = new Buffer(len)
+	const fd = fs.openSync(logpath, 'r')
+	fs.readSync(fd, buf, 0, len, startPos)
+
+	return cleanLog(buf.toString())
```
lukechampine, line 27: might be better to slice the buffer instead of the string. I doubt there's much performance impact though

lukechampine: `readSync` can throw an exception, right? Is it caught anywhere?

test/logs/testdir/gateway/gateway.log
```diff
@@ -0,0 +1,36 @@
...
+2016/11/16 22:07:34.127862 nodes.go:212: ERROR: could not pick a random node for uptime check: no peers
+2016/11/16 22:08:55.731417 gateway.go:245: STARTUP: Logging has started.
+2016/11/16 22:08:55.731520 gateway.go:343: INFO: gateway created, started logging
+2016/11/16 22:09:15.731846 nodes.go:212: ERROR: could not pick a random node for uptime check: no peers
+2016/11/16 22:09:18.868129 upnp.go:99: WARN: could not automatically forward port 9981: no UPnP-enabled devices found: no UPnP-enabled gateway found
+2016/11/16 22:09:25.648820 gateway.go:245: STARTUP: Logging has started.
+2016/11/16 22:09:41.101616 gateway.go:245: STARTUP: Logging has started.
+2016/11/16 22:09:41.101746 gateway.go:343: INFO: gateway created, started logging
+2016/11/16 22:10:01.102159 nodes.go:212: ERROR: could not pick a random node for uptime check: no peers
+2016/11/16 22:10:14.106560 upnp.go:83: INFO: our address is 76.119.238.92:9981
```
lukechampine: maybe mask the IP address

avahowell:
As of the latest commit, I've added more comprehensive testing, handled the exception from `readSync`, and changed the default button behavior to be a toggle.

avahowell:
Logs plugin is now hidden from the sidebar and can be activated by pressing `Control + Shift + L`. After some discussion and testing, the previous behavior of exclusively selecting logs on click and toggling them on ctrl-click or shift-click has been restored, and the buttons have been moved to the top.

PR #473, opened 2016-11-18
increase /renter/delete timeout to 5 minutes

avahowell:


lukechampine:
5 minutes may not be long enough, since it has to dial each of the hosts.
In the future, delete actions will be added to a queue (like upload currently is), so the call will be non-blocking.

DavidVorick:
(using a new API endpoint)

PR #481, opened 2016-12-08
replace instances of `fs` with `graceful-fs`

avahowell:
This PR completes Sia-UI's migration to `graceful-fs`.

### DIFF
--- a/js/mainjs/config.js
+++ b/js/mainjs/config.js
@@ -1,4 +1,4 @@
-import fs from 'fs'
+import fs from 'graceful-fs'
 import Path from 'path'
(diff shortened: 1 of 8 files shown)


DavidVorick:
Not sure if I missed a bunch of PRs but can you explain what the difference is?

DavidVorick:
nvm, I see that it's a drop-in replacement.

DavidVorick:
https://github.com/isaacs/node-graceful-fs

avahowell:
Yeah, there was a PR previous to this that explained the benefits. Definitely something we want.

PR #483, opened 2016-12-12
fix shift click not deselecting log filters

avahowell:
This PR fixes a bug that caused log filters not to be deselected when selected and shift-clicked.


lukechampine:
lint failed

avahowell:
fixed the linting problem

PR #485, opened 2016-12-13
compile the plugin api into each plugin instead of using `preload`

avahowell:
This PR fixes nondeterminism in the build system, where plugins could under certain circumstances on certain systems load before the plugin api, causing a 'white screen' crash. The plugin api is now compiled into each plugin using `webpack`, instead of preloaded.


lukechampine:
:+1:

PR #486, opened 2016-12-14
display 'Loading files...' instead of 'No files uploaded' before first api request

avahowell:
this pr changes the file list to show 'Loading files...' instead of 'No files uploaded' if an api response containing the user's files has not yet been received.

### DIFF
--- a/plugins/Files/js/components/filelist.js
+++ b/plugins/Files/js/components/filelist.js
@@ -72,21 +72,12 @@ const FileList = ({files, selected, searchResults, path, showSearchField, action
 		)
 	})
 
-	const fileListContents = (() => {
-		if (fileElements.size === 0) {
-			return (
-				<h2> No files uploaded </h2>
-			)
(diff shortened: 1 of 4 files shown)


plugins/Files/js/components/filelist.js
```diff
@@ -60,12 +60,27 @@ const FileList = ({files, selected, searchResults, path, showSearchField, action
...
 		)
 	})
+
+	const fileListContents = (() => {
+		if (!hasPolled) {
+			return (
+				<h2> Loading files... </h2>
+			)
+		}
+		if (fileElements.size === 0) {
```
lukechampine: would be nicer if we could just compare `fileElements === null` or `undefined` or something instead of needing a separate var.

avahowell: agreed. I changed the logic to return `Loading` if `files` is null.

PR #491, opened 2016-12-20
filter out uploads with <0 redundancy from uploads

avahowell:
This PR prevents files with -1 redundancy from being shown in the upload list.


plugins/Files/js/sagas/helpers.js
```diff
@@ -207,6 +207,7 @@ export const totalUsage = (files) => readableFilesize(files.reduce((sum, file) =
 // return a list of file uploads
 export const parseUploads = (files) =>
 	List(files)
+		.filter((file) => file.redundancy > 0)
```
lukechampine: shouldn't this be `>= 0`? IIRC files start with 0 redundancy

avahowell: yes, woops

PR #493, opened 2017-01-02
DRY production webpack config

avahowell:
this PR removes the duplication in our webpack production and development configs. The production config now imports the base config and adds the required production plugins.


lukechampine:
:+1: :+1:

PR #497, opened 2017-01-06
properly make Sia-UI single instance

### LINKED ISSUE #496: Users are able to launch multiple instances of Sia-UI
Despite leveraging electron's `app.makeSingleInstance()` method, users are still able to launch multiple instances of Sia-UI.

avahowell:
This PR utilizes `app.makeSingleInstance` properly, using its return value (bool) to prevent any extra instances of Sia-UI from running while an instance is already open. When a new instance is started, the original instance is restored and focused and the new instance is quit.

Fixes #496

### DIFF
--- a/js/mainjs/index.js
+++ b/js/mainjs/index.js
@@ -7,16 +7,24 @@ import initWindow from './initWindow.js'
 global.config = loadConfig(Path.join(__dirname, '../config.json'))
 let mainWindow
 
+// Allow only one instance of Sia-UI
+const shouldQuit = app.makeSingleInstance(() => {
+	if (mainWindow) {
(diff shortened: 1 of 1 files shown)


js/mainjs/index.js
```diff
@@ -7,18 +7,26 @@ import initWindow from './initWindow.js'
 global.config = loadConfig(Path.join(__dirname, '../config.json'))
 let mainWindow
 
+// Allow only one instance of Sia-UI
+const shouldQuit = app.makeSingleInstance(() => {
```
lukechampine, line 11: `alreadyExists` might be a better name

PR #500, opened 2017-01-12
use local shortcuts for dev tools and plugins

### LINKED ISSUE #499: Ctrl+Shift+P is a global hotkey
This prevents it from being used in other applications, e.g. SublimeText

avahowell:
This PR prevents the UI's shortcuts (Ctrl-Shift-L and Ctrl-Shift-P) from being persistently registered even when the UI is not focused. We still want to preserve the semantics of the `globalShortcut` api, namely parsing `Ctrl+Shfit+x` type shortcut values and being able to receive shortcut events regardless of what DOM element is focused, so this code simply unregisters shortcuts when focus is lost and re-registers them when focus is regained, instead of using the JS `window.onkeydown/keyup` api.

Fixes #499.

### DIFF
--- a/js/rendererjs/plugins.js
+++ b/js/rendererjs/plugins.js
@@ -33,6 +33,18 @@ const createPluginElement = (markupPath, title) => {
 	return elem
 }
 
+// registerLocalShortcut registers an electron globalShortcut that is only
+// active when the app has focus.
+const registerLocalShortcut = (shortcut, action) => {
+	remote.app.on('browser-window-blur', () => {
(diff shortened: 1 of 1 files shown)


lukechampine:
worth reading the discussion of this approach here: https://github.com/electron/electron/issues/1334#issuecomment-223857202

DavidVorick:
@lukechampine do you recommend that we not merge this PR then?

lukechampine:
It should be merged as long as it doesn't do anything wonky. Seems ok though.

PR #505, opened 2017-01-19
sort directories above files

avahowell:
This PR makes directories always sorted before files in the file list.

### DIFF
--- a/plugins/Files/js/sagas/helpers.js
+++ b/plugins/Files/js/sagas/helpers.js
@@ -70,6 +70,18 @@ export const minRedundancy = (files) => {
 	}).redundancy
 }
 
+// directoriesFirst is a comparator function used to sort files by type, where
+// the directories will always come first.
+const directoriesFirst = (file1, file2) => {
+	if (file1.type === 'directory' && file2.type === 'file') {
+		return -1
+	}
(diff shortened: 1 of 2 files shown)


plugins/Files/js/sagas/helpers.js
```diff
@@ -105,7 +105,19 @@ export const ls = (files, path) => {
 			type,
 		})
 	})
-	return parsedFiles.toList().sortBy((file) => file.name)
+	return parsedFiles.toList().sortBy((file) => file.name).sort(directoriesFirst)
```
avahowell, line 120: note that because two file types, `directory` and `file` happen to be lexicographically in the correct order for our desired behavior, we could do a `.sortBy((file) => file.type)` here, but I decided that this was too clever.

lukechampine: tfw to clever to write maintainable code

lukechampine:
Is `sort` a stable sort? If not, the files/folders may not remain lexicographically sorted after the second sort.

avahowell:
Yes, from the immutablejs docs:

```
sort()
Returns a new Iterable of the same type which includes the same entries, stably sorted by using a comparator.
```

PR #507, opened 2017-01-26
use downloads directory for siac in terminal plugin

avahowell:
This PR updates the terminal plugin to launch siac using Electron's `downloads` directory, useful to ensure that siac commands which export files go to a usable location.


lukechampine:
have you tested this locally?

DavidVorick:
would be nice to see this tested on both windows and linux

avahowell:
yes, I tested this locally. This relies on electron's `getPath` function (https://github.com/electron/electron/blob/master/docs/api/app.md#appgetpathname) which is guaranteed to work cross-platform.

PR #513, opened 2017-02-07
add a warning text next to balances when wallet is not synced

avahowell:
This PR adds a small, red warning text that appears under the balances in the wallet when it is not synced. This should help with the frequent 'Where is my money?' questions on the forum and elsewhere.

Screenshot:
notsynced


lukechampine:
LGTM

PR #514, opened 2017-02-10
fix parsing of files in file browser starting with '..'

avahowell:
This PR fixes an issue with parsing siapaths in the file browser: previously, if a user uploaded a file with '..' in its name, an extra non-existent directory would appear and navigating to this directory would break the file browser. I added a test to demonstrate this bug and slightly changed the `ls` and `onBackClick` algorithms to fix the bug.

### DIFF
--- a/plugins/Files/js/components/filelist.js
+++ b/plugins/Files/js/components/filelist.js
@@ -8,14 +8,20 @@ import DirectoryInfoBar from './directoryinfobar.js'
 
 const FileList = ({files, selected, searchResults, path, showSearchField, actions}) => {
 	const onBackClick = () => {
-		if (path === '') {
+		// remove a trailing slash if it exists
+		const cleanPath = path.replace(/\/$/, '')
+
+		if (cleanPath === '') {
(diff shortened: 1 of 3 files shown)


DavidVorick:
Was this the only major issue discovered?

DavidVorick:
also, any chance you got to try on multiple operating systems?

avahowell:
This was the first major issue discovered. I'm going through the list of naughty strings and trying strings that are also valid filenames on UNIX and Windows, will post subsequent PRs if those turn up issues.

I have not tried this on windows yet.

DavidVorick:
Are you not able to create an automated test which uses all the naughty strings?

avahowell:
I can, but not all naughty strings are valid filenames. I suppose that's not as important since siapaths do not have to be valid filenames, though

DavidVorick:
they don't? They end up on disk eventually, named matching their uploaded name. Is there an automated way you can strip out all the non-valid filenames from the BLNS?

PR #515, opened 2017-02-13
make closing behavior more obvious

avahowell:
Based on feedback on the forum and elsewhere, it has become apparent that the closing behavior of Sia-UI (closing to the task menu on windows, osx, and linux) is confusing for some users. This PR changes the behavior to minimize instead of hide the window, keeping it visible in the Windows taskbar. If the user then right clicks the Sia icon in the taskbar and clicks 'Close', a clean shutdown will be triggered. Sia-UI can still be quit from the task menu, but this should help users who are not used to that behavior and expect that the app is quit when it disappears from the taskbar.

### DIFF
--- a/js/rendererjs/index.js
+++ b/js/rendererjs/index.js
@@ -84,7 +84,7 @@ ipcRenderer.on('quit', async () => {
 let hasClosed = false
 window.onbeforeunload = () => {
 	if (window.closeToTray) {
-		mainWindow.hide()
+		mainWindow.minimize()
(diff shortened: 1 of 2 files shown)


lukechampine:
So when you click the X, it minimizes? That seems strange to me. Less strange on Mac I guess.
I figured the tray ballon ("Sia is still running!") would be sufficient, but apparently not.

avahowell:
It's actually fairly standard behavior. Skype and qBittorrent both do this (on Windows and Mac). The behavior is the same on linux as it was before

PR #521, opened 2017-02-28
add rescan warning message if loading takes longer than 30 seconds

avahowell:
This PR adds a warning message that appears if loading takes more than 30 seconds, explaining that startup can take a long time after upgrading due to rescans. The exact copy is

```
Loading can take a while after upgrading to a new version. Check the <a style="text-decoration: underline; cursor: pointer" id="releasenotelink">release notes</a> for more details.
```

### DIFF
--- a/js/rendererjs/loadingScreen.js
+++ b/js/rendererjs/loadingScreen.js
@@ -113,7 +113,7 @@ export default async function loadingScreen(initUI) {
 			overlayText.innerHTML= 'Loading can take a while after upgrading to a new version. Check the <a style="text-decoration: underline; cursor: pointer" id="releasenotelink">release notes</a> for more details.'
 
 			document.getElementById('releasenotelink').onclick = () => {
-				shell.openExternal('https://github.com/NebulousLabs/Sia/releases/tag/v1.1.1')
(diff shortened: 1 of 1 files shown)


js/rendererjs/loadingScreen.js
```diff
@@ -106,6 +106,18 @@ export default async function loadingScreen(initUI) {
...
 		return
 	}
+
+	// Set a timeout to display a warning message about long load times caused by rescan.
+	setTimeout(() => {
+		if (overlayText.textContent === 'Loading Sia...') {
+			overlayText.innerHTML= 'Loading can take a while after upgrading to a new version. Check the <a style="text-decoration: underline; cursor: pointer" id="releasenotelink">release notes</a> for more details.'
+
+			document.getElementById('releasenotelink').onclick = () => {
+				shell.openExternal('https://github.com/NebulousLabs/Sia/releases/tag/v1.1.1')
```
lukechampine: we can probably just use `/releases`. Would be a pain to update this link on each release; we might forget.

avahowell: yeah, good point

PR #522, opened 2017-03-12
make file list more compact

betaboot:
I changed some styles for the Files plugin to get a more compact file list.

compact_style

Would be great to place the text "50 contacts" below the allowance text so that the toolbar doesn't break when the File Transfer dialog is open. But that needs HTML changes which i did not want to touch...


avahowell:
this looks good to me. @lukechampine thoughts?

PR #523, opened 2017-03-14
display siad's stderr on loadingscreen if load fails

avahowell:
This PR improves the behavior of the UI's loading process, reading and displaying the STDERR and STDOUT from siad if siad closes unexpectedly. This should help our users by giving them more information as to why their `siad` failed to start.


DavidVorick:
This is much appreciated.

Issue #332, opened 2016-06-29
Allow a different quick-access password to be used instead of the full 28 word seed for unlocking the wallet


AdamBLevine:
I appreciate the need for security but it would be much better if I could use a seperate human memorable password to encrypt the local wallet instead of using the full seed.

lukechampine:
This is possible when using siac, but it's not supported (yet) when using Sia-UI.
To accomplish this, when you create a new wallet, do so with `siac wallet init -p`. You will then be prompted for an encryption password, distinct from the passphrase. You can then continue using this wallet from the UI as usual.

ANDREYCHEREPOV:
я скачал кошелёк а пароль из 28 символов не сохранил можно ли опять скачать будет новый пароль  если нет что делать?

DavidVorick:
@ANDREYCHEREPOV that really deserves its own issue. You should be able to load a seed from the wallet screen after you create a new wallet.

ANDREYCHEREPOV:
пароль для кошелька где его взять?

avahowell:
Closing, this was implemented in #526

PR #526, opened 2017-03-16
User settable passwords

avahowell:
This PR adds a long-requested feature by the community: the ability to initialize a wallet in the UI using a user-input passphrase. Addresses #332.

Screenshots:

sia1
sia2
sia3


lukechampine:
@DavidVorick we generally want to discourage people from supplying their own password, right? Perhaps instead of the design in this PR, we could make it a more difficult by having a little button or checkbox labeled "I want to use a custom password" that pops up the password entry field.

DavidVorick:
We can probably fix the font sizes too some we're here.

Yes the default should be to receive a seed, with an option to add your own password instead if you hit a button.

avahowell:
ACK, I will change the design to prioritize using a seed.

avahowell:
Updated this PR to add a small checkbox under the 'create wallet' button, which enables the password form. If this button isn't checked, the wallet is initialized in the usual way. I also upped the font size for the seed and the password.

Screenshots:
sia1
sia2

lukechampine:
lgtm although I always get a bit unnerved when I see my password printed in plaintext. Not sure how @DavidVorick feels about it, but I think I'd prefer to have it replaced with ***** or something

DavidVorick:
Maybe display the first two and last one character? 

Also, I don't think we have an API test for creating a wallet with the encryption password. Before we merge this, can we merge a test in the Sia repo that:

1. Creates a wallet with an encryption password
2. Unlocks that wallet and uses it
3. Reboots
4. Unlocks the wallet and uses it
5. Makes sure the seed can be loaded into a new wallet correctly

lukechampine:
@johnathanhowell we're thinking that something like this is best:

Security-wise it's kinda silly because the seed is in plaintext and that's much more important to protect, but seeing your password in plaintext is just unnerving.

avahowell:
What do you think about displaying the password with a simple spoiler tag-style interface? (it shows up as a black line of text until you hover)

lukechampine:
that sounds fine to me

avahowell:
done

PR #530, opened 2017-03-27
add host connectability component

avahowell:
This PR adds a new component to the Host plugin's header which displays the connectability status of the host's `NetAddress`.

Screenshots:
host2
sia1


DavidVorick:
'Host Connectable' is pretty clunky for terminology. Also, probably does not need to be displayed if the host is both not accepting contracts and also does not have any contracts (it means the user is not running a host).

Fornax96:
The 20000 number is more of an advice than a hard requirement right? What if the price triples tomorrow, will the UI need to be updated with a new number?

RE: 'Host Connectable' is pretty clunky for terminology.
I'd say "Host Reachable" sounds better.

plugins/Hosting/js/sagas/saga.js
```diff
@@ -204,6 +206,33 @@ function *requestDefaultSettingsSaga() {
...
 }
 
+// getHostStatus returns a promise that resolves with `true` if the provided
+// netaddress is connectable, or `false` if it is not connectable. The input
+// 'hostAddr' must be an address in the URL format, eg:
+// tcp://123.456.789.0:9982
+const getHostStatus = (hostAddr) => new Promise((resolve) => {
+	const hostUrl = url.parse(hostAddr)
+	const socket = net.connect(hostUrl.port, hostUrl.hostname)
+	socket.setTimeout(15e3) // 15 second host connectability timeout
```
DavidVorick: I'd aim for closer to 10 minutes. Host status isn't going to change faster than that, and it gives you more time to be certain that you are actually offline if nobody is interacting with you.

DavidVorick: oh wait, this doesn't use the RPC metrics. I see now why you say 'connectable'.

If you look at the RPC metrics you will be able to tell whether people are actually connecting to you or not. Just because you are connectible doesn't mean that your most recent host announcement has the right net address, or that there isn't some other misconfiguration that is preventing renters from seeing you. And after we add proof of burn, you might just not have burned enough coins to be grabbing renter's attention.

So you really do want to be looking at the RPC metrics, they are going to give you a more complete picture than simply pinging yourself.

lukechampine: You would need to save the initial values on startup and then check for an increase in non-error calls. Not sure how long it takes on average to get one connection.
I think having both would be ideal, but make it clear that one of them just means your port is properly opened. As an example, Transmission has a button that you can use to manually test your port:

DavidVorick: Less than 10 minutes to get your first ping.

Interestingly, seems that hosts can get pings even if their ports are not open. Presumably from their own hostdb? Not sure

avahowell: I'm in favor of having both an open port check component and a more fuzzy 'Host Health' component, I believe both have their uses. It's useful to know if your Host is connectable from the outside world separately from if you are receiving settings calls. What if we add another component to this PR that conveys a 'health' metric based on the historic values of the RPC metrics?

DavidVorick: That's probably too abstract. If it's not immediately obvious what the thing means, we're going to have 100 users asking us what it means and why it's not 100%, etc.

I think what would be best would be to have a single field in the bar, like we already have, that only displays if you are actively hosting (or have contracts), which either says 'Host is Working' (or something to that effect), 'Host cannot be Reached' (or something to that effect), or 'Host is not Visible to Renters' (or something to that effect). And if we add more potential ways to detect that something in the host does not seem to be working, we'll add those too to the potential failure messages.

And if multiple failures are active, we'll have a sensible priority. For example, if your ports aren't forwarded, then it doesn't really make sense to point out that renter's can't contact you. Just mentioning the port thing should be enough.

avahowell: Agreed- I'll update this PR to add a few more states for this component, based on the Host's RPC stats.

avahowell:
I've updated the host status component to track both the connectability and the 'working status' of the host, where 'working status' is determined by if the host's `SettingsCalls` metric has increased over the period of 15 minutes. The component displays 3 states:

- Host Unreachable (cannot be connected to at its NetAddress)
- Host Inactive (can be connected to, but has not received a SettingsCall in the last 15 muntes)
- Host Online (is both connectible and has received a settingscall in the last 15 minutes).

I also added a helpful hover arrow box that explains each state.

Screenshots:

host1
host2
host3
host5

plugins/Hosting/js/sagas/saga.js
```diff
@@ -204,6 +206,63 @@ function *requestDefaultSettingsSaga() {
...
+	})
+})
+
+// hostWorkingStatusSaga compares the host's current `settingscalls` metric and
+// sets `working` to false if it has not increased.
+function *hostWorkingStatusSaga() {
+	try {
+		const hostData = yield siadCall('/host')
+		const nPreviousSettingsCalls = yield select((state) => state.hostingReducer.get('nSettingsCalls'))
+		if (nPreviousSettingsCalls >= hostData.networkmetrics.settingscalls) {
```
DavidVorick: I would look, over the course of 15 minutes, for at least 3 settings calls, instead of just 1

avahowell:
I added an additional state for this component, displayed on launch before the first check happens:

host6
host7

PR #532, opened 2017-04-01
add progress bar to sync status component

avahowell:
This PR adds a progress bar to the sync status component, to inform users the progress of IBD.

sia1

### DIFF
--- a/js/rendererjs/statusbar.js
+++ b/js/rendererjs/statusbar.js
@@ -15,9 +15,9 @@ const currentEstimatedHeight = () => {
 }
 
 // estimatedProgress returns the estimated sync progress given the current
-// blockheight, as a number from 0 -> 100
+// blockheight, as a number from 0 -> 99
 const estimatedProgress = (currentHeight) =>
-	currentHeight / currentEstimatedHeight() * 100
+	Math.min(currentHeight / currentEstimatedHeight() * 100, 99)
 
 // -- components --
 
--- a/js/rendererjs/statusbar.js
+++ b/js/rendererjs/statusbar.js
@@ -1,14 +1,57 @@
 import React, { PropTypes } from 'react'
 
-const redColor = '#E0000B'
-const greenColor = '#00CBA0'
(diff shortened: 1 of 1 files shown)


js/rendererjs/statusbar.js
```diff
@@ -1,14 +1,59 @@
...
+
+	const estimatedHeight = knownBlockHeight + (diffMinutes / blockTime)
+
+	return Math.floor(estimatedHeight + 0.5) // round to the nearest block
 }
 
+// estimatedProgress returns the estimated sync progress given the current
+// blockheight, as a number from 0 -> 100
+const estimatedProgress = (currentHeight) =>
+	currentHeight / currentEstimatedHeight() * 100
```
lukechampine: In siac we explicitly avoid displaying 100% progress, instead capping at 99%. Should probably do the same here.

js/rendererjs/statusbar.js
```diff
@@ -15,9 +15,9 @@ const currentEstimatedHeight = () => {
 }
 
 // estimatedProgress returns the estimated sync progress given the current
-// blockheight, as a number from 0 -> 100
+// blockheight, as a number from 0 -> 99
 const estimatedProgress = (currentHeight) =>
-	currentHeight / currentEstimatedHeight() * 100
+	currentHeight / currentEstimatedHeight() * 99
```
lukechampine: unfortunately I think this won't quite do it, because `currentEstimatedHeight()` is just that -- an estimate. It's possible for currentHeight to exceed `currentEstimatedHeight()`, which would push this into 100% again. So we need an explicit branch for `if > 99 then 99`

avahowell: Yeah, you're right

avahowell: Don't need a branch though, we can use `Math.min` here

lukechampine: the things you miss when using Go... :cry:

PR #543, opened 2017-04-07
add 'no file transfers in progress' header if there are no file transfers

avahowell:


lukechampine:
build failed?

avahowell:
looks like there's some non-determinism in the build, rebased and force pushed and it's passing now, as is the other PR which failed on the same tests

PR #546, opened 2017-04-07
round storage folder size to the nearest 256MB

avahowell:


plugins/Hosting/js/sagas/saga.js
```diff
@@ -68,6 +68,14 @@ function *addFolder(action) {
 	}
 }
 
+// bytesToStorageGB converts a number of bytes to a valid size of storage, in
+// GB rounded to the nearest 256MB, as a BigNumber.
+const bytesToStorageGB = (bytes) => {
+	const remainder = bytes % 256e8
+	const roundedBytes = bytes + 256e8 - remainder
```
lukechampine: @DavidVorick I believe it's 256MiB, not 256MB, right?

DavidVorick: Yes, MiB.

DavidVorick:
Best to round down, not to nearest.

avahowell:
this is tested and ready to merge now

plugins/Hosting/js/sagas/saga.js
```diff
@@ -52,14 +52,25 @@ function *announceHost(action) {
 	}
 }
 
+// bytesToStorage converts a BigNumber of GB to a valid size of storage, in
+// bytes, rounded to the nearest 256MiB (64 * SectorSize).
+const bytesToStorageBytes = (bytes) => {
+	const roundedBytes = bytes.minus(bytes.modulo(2.6843546e8))
+	if (roundedBytes.isNegative()) {
+		return '0'
+	}
```
lukechampine, line 61: is it even possible for this to be negative if `bytes > 0`?

plugins/Hosting/js/sagas/saga.js
```diff
@@ -52,14 +52,25 @@ function *announceHost(action) {
 	}
 }
 
+// bytesToStorage converts a BigNumber of GB to a valid size of storage, in
+// bytes, rounded to the nearest 256MiB (64 * SectorSize).
+const bytesToStorageBytes = (bytes) => {
+	const roundedBytes = bytes.minus(bytes.modulo(2.6843546e8))
```
lukechampine, line 58: kind of a weird number. Might as well just write out `268435456` or `64 * 4194304`

plugins/Hosting/js/sagas/saga.js
avahowell: it turns out that it is not possible for this to be negative, since the UI enforces a minimum size of 35GB in the dialogs. I wanted to program this as defensively as possible, though.

plugins/Hosting/js/sagas/saga.js
avahowell: It's also an option to pull the `SectorSize` from the API and figure out the math that way, which would be more robust. Perhaps overkill, though (and it would require changing this function to a promise or an `async` function, since the API request would be asynchronous)

PR #547, opened 2017-04-07
add init from seed to uninitialized wallet screen

avahowell:
This PR adds the ability for users to initialize their wallet from a seed, on the uninitialized wallet screen
sia1
sia2
sia3
sia4


DavidVorick:
If you init from seed, do you need a password, or are you setting a password?

avahowell:
You can optionally provide a password (I tried to show that behavior in the screenshots) using the checkbox

DavidVorick:
Ah. Text should be 'Desired Password'

avahowell:
/init/seed can take a long time, this needs longer timeouts and some status feedback. Working on that now

avahowell:
Should be ready now

lukechampine:
Looks good but perhaps could use some more description. Users should understand that init-from-seed has to scan the whole blockchain, so it's slow and requires you to be synced.

avahowell:
updated with some text: 
sia1

PR #555, opened 2017-04-13
sort storage folders by name instead of size

### LINKED ISSUE #554: Host storage folders still reordering
v1.2.0-RC3

Me again :)
Storage folders are now sorted by size, but i have two folders with the same size and so they would still swap places from time to time. Could the list be sorted by name (i think siac is doing that)?

avahowell:
fixes #554


lukechampine:
oops, should have caught this during my review

PR #560, opened 2017-04-14
dont launch siad with miner module

avahowell:
Nodejs-Sia's `launch` function uses the miner module by default, this PR changes the UI to only launch using consensus, gateway, host, renter, tpool, and wallet modules.


lukechampine:
can you open a PR to fix nodejs-sia?

avahowell:
Yeah, what is the desired behavior? Use the daemon's module defaults?

lukechampine:
that seems most sensible to me. It should just drop the modules flag.

PR #561, opened 2017-04-14
try the default siad path if it cannot be located on load

avahowell:
This PR makes the loadingscreen a bit smarter, it will now try the default siad path (`Sia/siad`) if the path loaded from `config.json` does not exist. This solves an issue where users who rename `Sia-UI.app` on Mac are prompted to relocate `siad`.


DavidVorick:
LGTM

PR #569, opened 2017-04-21
only display 'Host Unreachable' if not connectable and not working

avahowell:
Previously `Unreachable` would be displayed if the host was not connectable, regardless of the working state.


DavidVorick:
LGTM

PR #571, opened 2017-04-21
add min height to confirmation dialog

### LINKED ISSUE #1728: When creating new wallet; OK-button hidden behind cancel-button
Sia 1.2.0 running on Windows 7. After it asks me to write down the seed and passphrase it asks me to enter the passphrase again to unlock the wallet. Below the input field is a cancel button. Nothing else. I try to hit enter but it is ignored. When I hit TAB I see what's going on. There is an OK button hidden behind the cancel button. So you can proceed by entering the passphrase, then hitting TAB and finally enter.

In addition to ensuring the OK button is visible, it would be nice if it was possible to just hit enter after typing the passphrase.

I was running on low resolution when this happened, so it may not happen to a lot of people. I found it strange though as there was plenty of unused space for the OK button.

avahowell:
fixes https://github.com/NebulousLabs/Sia/issues/1728

### DIFF
--- a/plugins/Wallet/css/wallet.css
+++ b/plugins/Wallet/css/wallet.css
@@ -246,6 +246,7 @@
 	background-color: #ECECEC;
 	width: 75%;
 	height: 75%;
+	min-height: 410px;
 }
 .newwallet-dialog button {
 	margin-top: 20px;


lukechampine:
huh. I'm surprised that GitHub will auto-close issues across repos. Maybe it's because I'm a maintainer of both?

DavidVorick:
heh. That would be a really fun vulnerability.

Close 4 million issues across 100,000 software projects in a single PR xD

PR #572, opened 2017-04-24
replace electron version in OSX .app in build script

avahowell:
This PR adds version substitution to the existing .plist modification in the build script, replacing the Electron version with `siaVersion`.


lukechampine:
what is the effect of this?

avahowell:
Applications that read version info from the `.app` bundle, such as Finder (if you Ctrl-I) and other system utilities, will have the correct version.

PR #573, opened 2017-04-27
Removed maxRedundancy

lolsteve:
Since changing the maximum redundancy, newly uploaded files (with full redundancy) would show some opacity.

screen shot 2017-04-26 at 9 20 35 pm

screen shot 2017-04-26 at 9 48 54 pm


lolsteve:
Actually, I'm not sure if this shows the correct information. It will be fully opaque once the redundancy is at 1x. The opaqueness should be a reflection of the progress.

Issue #395, opened 2016-09-27
Sia 1.0.1 | Sia UI v1.0.2 | Mac OSX sierra requires sudo to open.


imjackofitall:
Getting the following error when opening Sia UI on Mac OSX Sierra:

![screen shot 2016-09-27 at 11 39 08](https://cloud.githubusercontent.com/assets/2267993/18857422/19ff37e6-84a7-11e6-89d6-253621f13663.png)

App runs fine when running sudo Sia-UI.app/Contents/MacOS/Sia-UI

Run a few checks on my end and came up short, is there some sort of permission issue?

lukechampine:
See https://github.com/NebulousLabs/Sia/issues/1441
tl;dr: OSX is invisibly moving the app to a read-only temporary folder because it doesn't trust it. Moving the file to a different location should prevent this behavior.

We might need to put this in a FAQ or something. I think it only happens with the latest OSX release.

DavidVorick:
Is there potentially any OS specific code we can write to prevent this?

avahowell:
IIRC the correct way to prevent this is to enroll in Apple's dev program and sign our release binaries.

DavidVorick:
Well, that's not too terrible once we're further along. I guess we can put it on the 'eventually, but sooner eventually rather than later eventually' list.

imjackofitall:
An easy fix would be to have the dmg open up with the .app and a link to the applications folder there. Would be a lot more effective than just a amendment to the FAQ.

DavidVorick:
Re-opening this issue as there seem to be action items still up for discussion

lukechampine:
Fixed in #578

PR #578, opened 2017-05-01
move config.json to userData

avahowell:
this PR moves `config.json` to `userData` (the same location where Sia's data is stored), instead of inside the app bundle. This will preserve UI settings across upgrades but more importantly it means that `Sia-UI.app` will run perfectly fine on read-only filesystems, solving #395 and making it so users do not have to move the app before using it.

This has been tested using the latest Mac OS X Sierra.

### DIFF
--- a/js/mainjs/index.js
+++ b/js/mainjs/index.js
@@ -4,7 +4,7 @@ import loadConfig from './config.js'
 import initWindow from './initWindow.js'
 
 // load config.json manager
-global.config = loadConfig(Path.join(__dirname, '../config.json'))
+global.config = loadConfig(Path.join(app.getPath('userData'), 'config.json'))
 let mainWindow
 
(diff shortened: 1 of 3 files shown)


lukechampine:
`config.json` was the only thing causing that bug???

avahowell:
correct, `config.json` saves when it is loaded and this requires write access, app translocation moves the app bundle to a read-only filesystem

lukechampine:
Build is consistently failing with:
```
  1) startup and shutdown behaviour window closing behaviour quits gracefully on close if closeToTray = false:
     TypeError: Cannot read property 'map' of null
      at node_modules/webdriverio/build/lib/protocol/elements.js:106:36
      at process._tickDomainCallback (internal/process/next_tick.js:129:7)
```
do we need to update one of our dependencies?

PR #587, opened 2017-05-10
don't display the lockscreen if we are initializing a seed

avahowell:
This fixes the issue where users could see a lockscreen and attempt to unlock during seed initialization.


DavidVorick:
Would like to see unit tests accompanying these PRs. That's how we make sure we don't get regressions, on top of just verifying it works in the first place.

avahowell:
This PR included a unit test: https://github.com/NebulousLabs/Sia-UI/pull/587/commits/bb3d0073875ffd47c8c0645f6f0b0b34c5ee6381#diff-3f1974fdfc963ead85c7f3af84fc4129R23

DavidVorick:
Derp I am just blind. Thanks

PR #590, opened 2017-05-12
Use babel preset env

moimael:
It is the recommended preset to use currently. It will only transpile what is needed to make it work with the version of electron we use, lowering transpile time and improving performances.

Also remove the duplicate babel-loader query in webpack since we already use a .babelrc which is the recommended way to configure babel.

Moves babel-preset react and stage-3 to devDependencies.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,11 @@
 {
-  "presets": ["es2015", "react", "stage-3"]
+  "presets": [
+    ["env", {
+      "targets": {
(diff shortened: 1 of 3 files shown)


avahowell:
Thanks for this PR, I've been thinking about doing this and updating the other dependencies in the stack but haven't gotten around to it yet. I pulled and verified these changes yield pretty large compilation time speedups and don't break our build script, so this looks good to go.

PR #593, opened 2017-05-15
Split vendor dependencies in a common chunk

moimael:
Simple implementation of CommonChunk webpack plugin that moves npm dependencies into vendor.js. Works really well with the current setup.

In the future we might want to have a vendor js file per plugin and a meta vendor js file that removes duplicates across them.

Also extracted webpack loader code into manifest.js. It allows to recompile less often in dev, since this code almost never changes.

Size comparaison:
Before (**7.49 MB, compile time: 178666ms**):
```
> webpack --config ./webpack.production.config.js

Hash: 7256b838e5c13ef5d13a
Version: webpack 2.5.1
Time: 178666ms
                Asset     Size  Chunks                    Chunk Names
   ./plugins/Files.js  1.13 MB       0  [emitted]  [big]  ./plugins/Files
  ./plugins/Wallet.js  1.11 MB       1  [emitted]  [big]  ./plugins/Wallet
 ./plugins/Hosting.js   1.1 MB       2  [emitted]  [big]  ./plugins/Hosting
./plugins/Terminal.js  1.07 MB       3  [emitted]  [big]  ./plugins/Terminal
    ./plugins/Logs.js  1.06 MB       4  [emitted]  [big]  ./plugins/Logs
          renderer.js  1.05 MB       5  [emitted]  [big]  renderer
   ./plugins/About.js   974 kB       6  [emitted]  [big]  ./plugins/About
              main.js  14.3 kB       7  [emitted]         main
```

After (**1.25 MB, compile time: 62092ms**):
```
> webpack --config ./webpack.production.config.js

Hash: 1b8c253826adabe9ba4ddcb187723debb0b4eda8
Version: webpack 2.5.1
Child
    Hash: 1b8c253826adabe9ba4d
    Time: 12195ms
      Asset     Size  Chunks             Chunk Names
    main.js  14.2 kB       0  [emitted]  main
       [2] ./~/graceful-fs/fs.js 429 bytes {0} [built]
       [4] ./js/mainjs/index.js 967 bytes {0} [built]
       [5] ./js/mainjs/appMenu.js 1.02 kB {0} [built]
       [6] ./js/mainjs/config.js 1.83 kB {0} [built]
       [7] ./js/mainjs/initWindow.js 1.92 kB {0} [built]
       [8] ./js/mainjs/trayMenu.js 418 bytes {0} [built]
       [9] ./~/graceful-fs/graceful-fs.js 6.94 kB {0} [built]
[...]

PR #595, opened 2017-05-15
Add babel caching

moimael:
Reduce recompile time in dev mode from **26472ms** to **16784ms** on my computer.

Based on #593


avahowell:
Is there any reason not to use caching? (unexpected side effects like sources not being rebuilt, etc)

moimael:
That could happen, as always with caching, but I never had any issues with it and I'd say from my experience, that it is a very safe option to enable (we are using it in a fairly large js app at work). And if there are issues, you can change the cache identifier to use other parameter for detecting a needed rebuild.

PR #597, opened 2017-05-15
Use system fonts

moimael:
This is the configuration used by Atom and also medium before. It will use the system font on Ubuntu, Gnome, KDE, Windows, MacOS.

Remove the Roboto font except for Terminal which uses the monospace variant.


lukechampine:
can you provide screenshots?

moimael:
Here is a screenshot on gnome:
https://pasteboard.co/72qfK22hF.png

avahowell:
I like this change, but it might be controversial. @DavidVorick @lukechampine thoughts?

lukechampine:
hmm. I like having a consistent look across all platforms. What's the main impetus for using system fonts?

moimael:
It looks more like a system app and less like a website. But as johnatanhowell stated, this is a bit of a controversial topic. I just know that the first time I launched the app, it looked weird to me because of that.
Atom is doing that too, and I wasn't able to check but I guess slack is doing that too. Both apps look very sleek font wise on Linux and feel more like an actual app

PR #599, opened 2017-05-16
Add autoprefixer.

moimael:
Remove vendor prefixes and use a browserlist to state which target we support.
Soon babel-preset-env will support the browserlist file too so we can simplify the configuration further.

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,0 +1,1 @@
+electron 1.6
--- a/css/general.css
+++ b/css/general.css
@@ -29,20 +29,20 @@ html,body {
 	font-weight: 300;
 	font-size: 18px;
 
-	-webkit-user-select: none;
+	user-select: none;
 }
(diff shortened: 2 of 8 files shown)


lukechampine:
Is there reason to expect that we'll need to support a non-webkit target in the future? Just unsure if this makes things more clear or less clear if you're a new dev diving into the codebase. Typically we try to keep our dependencies as minimal as possible.

moimael:
It's not really for non webkit support in this case, more for not yet standard props. The problem with prefixed css properties is that they are forgotten and will never get changed, while the standard one might get some improvement from browser engine dev (see flexbox for example). With autoprefixer you only write standard css and it will add prefixes when needed and remove them when they are fully supported.

Given the current size of the project, the benefit is not obvious, but when/if it grows, I think this is really worth it.

Also the dependency is only for dev, so it wont make the program larger and is quite small with minimal configuration.

PR #600, opened 2017-05-16
Remove unused variants of font awesome

moimael:
Electron only needs woff2


avahowell:
this doesn't break our icons right?

moimael:
Nope just removing the fonts that were here don't old browser compatibility.

Fontawesome tries woff2 first since it's the most modern format. Tried it and all icons are still there :)

PR #603, opened 2017-05-16
add support for `rescanning` to Wallet UI

avahowell:
This PR adds support for the new `rescanning` bool returned by the wallet API to indicate if an active rescan is occuring. If it is, the lockscreen is brought up with a 'Scanning the Blockchain...' dialog, informing users of the rescan and preventing wallet actions until the rescan is complete. This improves UX, users will now see what is happening more clearly (example: Unlock -> "wallet is unlocking..." -> "scanning the blockchain..." -> complete, same applies to initializing from a seed and sweeping a seed, even if those actions are performed outside of the UI).

I updated the UI's unit tests and performed end-to-end tests for this PR using Sia-Antfarm, verifying that the desired behavior occurs.

Screenshot of the 'Scanning' dialog:

sia2


plugins/Wallet/js/reducers/wallet.js
```diff
@@ -82,6 +83,21 @@ export default function walletReducer(state = initialState, action) {
...
+	case constants.SET_RESCANNING: {
+		let newstate = state.set('rescanning', action.rescanning)
+		// the `rescanning` state is mutually exclusive with `recovering`,
+		// `initializingSeed`, and `unlocking`. This makes sense for a user, they
+		// will see 'recovering' or 'initializing' change to 'rescanning', an
+		// accurate representation of what is happening on the backend.
+		if (action.rescanning) {
+			newstate = newstate.set('recovering', false)
+			                   .set('initializingSeed', false)
+			                   .set('unlocking', false)
```
lukechampine, line 95: if we add any more of these, may make more sense to use a single field with a constant (likely a string) instead of separate bools.

PR #617, opened 2017-05-29
add Change Password functionality to wallet plugin

avahowell:
This PR leverages the new `/wallet/changepassword` endpoint, allowing users to change their encryption password from the Wallet plugin UI.


avahowell:
Screenshots:
changepw1
changepw2
changepw3

PR #630, opened 2017-06-13
removed unnecessary args in checkSiaPath

eddiewang:
The siadConfig.path was being passed into a Promise that doesn't take arguments. Should work with it removed.


avahowell:
5000+ diff lol, remove package-lock.json pls

eddiewang:
Added the gitignore line as well

lukechampine:
Please rebase c823df3 so that package-lock.json does not appear in the commit history.

eddiewang:
I must have screwed something up. I'll just make another branch.

PR #640, opened 2017-06-22
fix config.json upgrading issue

avahowell:
This PR updates `config.json` by adding a version field. If this version doesn't exist, or is less than the current UI version, the `siad` specified in `config.json` will be overwritten to Sia-UI's new Siad. This is required since we moved `config.json` to the data directory, where it is reused across upgrades.

### DIFF
--- a/js/mainjs/config.js
+++ b/js/mainjs/config.js
@@ -1,6 +1,8 @@
 import fs from 'graceful-fs'
 import Path from 'path'
 import { app } from 'electron'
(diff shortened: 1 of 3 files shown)


lukechampine:
Why do we override the siad path if the version is older? Wouldn't we want to preserve that across upgrades?

avahowell:
No, because then the new Sia-UI release will use the old `siad`. If the version is older, we want to replace `siad` with the newly packaged `siad`.

PR #644, opened 2017-06-26
add fee estimation to wallet send prompt

avahowell:


plugins/Wallet/js/sagas/wallet.js
```diff
@@ -222,6 +222,16 @@ function *changePasswordSaga(action) {
 	}
 }
 
+
+function *startSendPromptSaga() {
+	try {
+		const response = yield siadCall('/tpool/fee')
+		const feeEstimate = SiaAPI.hastingsToSiacoins(response.maximum).times(1e3).round(8).toString() + ' SC/TB'
```
lukechampine: SC/TB?

The response here is (I believe) hastings per byte, which makes this tricky since we also need to estimate how many bytes the transaction will require. Would be good to have concrete numbers on this, but `1e3` is probably fine for now.

avahowell: This should be SC/KB, amending now

Issue #648, opened 2017-06-29
Files Tab UI nearly entirely unresponsive


Zelex:
If I click and wait a couple hours it eventually does something. 

I think this is due to some kind of O(n^2) or something regarding the number of files.

avahowell:
Fixed in #654

PR #654, opened 2017-07-03
optimize file parsing, fix directory navigation bug

avahowell:
This PR optimizes the UI's file parsing, fixing a case where the entire file tree would be traversed unnecessarily `n` amount of times per directory. It also fixes a bug where the back button would not work correctly with file paths deeper than 3 directories deep. I added a regression test for the bug, and profiled the file parsing changes manually. This should help users who experience large slowdowns with lots of uploaded files.


lukechampine:
Is this the suspected cause of #648?

avahowell:
Yes.

avahowell:
FYI, the profiling results are quite clear:

Without this optimization, with a directory tree a bit over a thousand files:
profile1

With the optimization:
sia2

PR #660, opened 2017-07-05
add `New Folder` button

avahowell:
This PR adds a 'New Folder' button to the files plugin toolbar, enabling users to create folders in the file list.


plugins/Files/js/components/uploadbutton.js
```diff
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types'
 import React from 'react'
 
-const minimumContracts = 14
+const minimumContracts = 2
```
lukechampine: ?

avahowell: leftover from testing, oops

PR #663, opened 2017-07-10
strip leading and trailing whitespace from seed, password

### LINKED ISSUE #662: Strip leading and trailing whitespace from seed
Sia-UI will reject a seed if it contains leading or trailing whitespace. Can the UI automatically strip leading or trailing whitespace since it knows that can't really be part of the seed? It's easy for users to accidentally include leading or trailing whitespace if they've saved the seed to a text file and are manually copy/pasting.

avahowell:
fixes #662


lukechampine:
To go along with this, we might want to prohibit users from setting passwords with leading or trailing whitespace (or at least warn them with a scary pop-up)

PR #664, opened 2017-07-10
Update checker

lolsteve:
Added update checker button to `About` plugin. Uses built in `siad` update check then forms a download link with the version given and value returned from `os.platform()`.

Initial state:
deepinscreenshot_select-area_20170710162940

After checking for update:
deepinscreenshot_20170710163012

If already updated:
deepinscreenshot_20170710163106

### DIFF
--- a/plugins/About/js/index.js
+++ b/plugins/About/js/index.js
@@ -1,6 +1,5 @@
 'use strict'
 import { platform } from 'os'
-import { format } from 'util'
 
 // Set UI version via package.json.
 document.getElementById('uiversion').innerHTML = VERSION
@@ -15,13 +14,13 @@ SiaAPI.call('/daemon/version', (err, result) => {
 	}
 })
 
-function genDownloadLink(version, platform) {
-	let plat = platform
+function genDownloadLink(version, thePlatform) {
+	let plat = thePlatform
 	if (plat === 'darwin') {
 		plat = 'osx'
 	}
 
-	return format('https://github.com/NebulousLabs/Sia-UI/releases/download/v%s/Sia-UI-v%s-%s-x64.zip', version, version, plat)
+	return `https://github.com/NebulousLabs/Sia-UI/releases/download/v${version}/Sia-UI-v${version}-${plat}-x64.zip`
 }
 
 function updateCheck() {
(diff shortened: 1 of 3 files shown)


plugins/About/js/index.js
```diff
@@ -13,3 +15,30 @@ SiaAPI.call('/daemon/version', (err, result) => {
...
+	}
+
+	return util.format('https://github.com/NebulousLabs/Sia-UI/releases/download/v%s/Sia-UI-v%s-%s-x64.zip', version, version, plat)
+}
+
+function updateCheck() {
+	SiaAPI.call('/daemon/update', (err, result) => {
+		if (err) {
+			SiaAPI.showError('Error', err.toString())
+			ipcRenderer.sendToHost('notification', err.toString(), 'error')
```
avahowell: `ipcRenderer` and notifications are deprecated

plugins/About/js/index.js
```diff
@@ -1,4 +1,6 @@
 'use strict'
+const os = require('os')
```
avahowell: `import`

plugins/About/js/index.js
lolsteve: Anything I should replace it with or just remove?

plugins/About/js/index.js
```diff
@@ -21,7 +21,7 @@ function genDownloadLink(version, platform) {
 		plat = 'osx'
 	}
 
-	return util.format('https://github.com/NebulousLabs/Sia-UI/releases/download/v%s/Sia-UI-v%s-%s-x64.zip', version, version, plat)
+	return format('https://github.com/NebulousLabs/Sia-UI/releases/download/v%s/Sia-UI-v%s-%s-x64.zip', version, version, plat)
```
avahowell:
```
return `https://github.com/NebulousLabs/Sia-UI/releases/download/v${version}/Sia-UI-v${version}-${plat}-x64.zip`
```
:)

plugins/About/js/index.js
avahowell: just remove, there's nothing that currently duplicates this functionality

DavidVorick:
Did we make that button look prettier than in the screenshot?

PR #671, opened 2017-07-13
add support for dragging/dropping folders

avahowell:
This also adds support for dragging into the 'back' button to move files and folders into the parent directory.


plugins/Files/js/sagas/files.js
```diff
@@ -235,7 +242,7 @@ function* renameFileSaga(action) {
 				},
 			})
 			yield put(actions.getFiles())
-		}
+		} else
 		if (action.file.type === 'directory') {
```
lukechampine: should be on same line, right?

PR #681, opened 2017-07-21
add deep search functionality

avahowell:
This PR changes the search functionality to search for all files and folders under the current directory matching the search text. Previously, only the filenames in the current directory would show up.

### DIFF
--- a/plugins/Files/js/sagas/helpers.js
+++ b/plugins/Files/js/sagas/helpers.js
@@ -219,9 +219,35 @@ export const parseUploads = (files) =>
 
 // Search `files` for `text`, excluding directories not in `path`
 export const searchFiles = (files, text, path) => {
-	let matchingFiles = List(files).filter((file) => file.siapath.indexOf(path) !== -1)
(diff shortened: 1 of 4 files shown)


plugins/Files/js/sagas/helpers.js
```diff
@@ -219,9 +219,35 @@ export const parseUploads = (files) =>
...
+	  .filter((file) => file.siapath.indexOf(path) === 0 && file.siapath !== path)
+	  .filter((file) => file.siapath.toLowerCase().includes(text.toLowerCase()))
+
+	let parsedFiles = Map()
+	filteredFiles.forEach((file) => {
+		let type = 'file'
+		let name = Path.posix.basename(file.siapath)
+		let siapath = file.siapath
+		const pathComponents = file.siapath.split('/')
+		if (!pathComponents[pathComponents.length - 1].toLowerCase().includes(text.toLowerCase()) || file.siaUIFolder) {
```
lukechampine: isn't `pathComponents[pathComponents.length - 1]` the same as `Path.posix.basename(file.siapath)`?

avahowell: Yeah, I think so

PR #684, opened 2017-07-30
Inspector

tlightsky:
this pr aimed to provide more detail of files in ui:
image

image
image

must work with this siad pr:https://github.com/NebulousLabs/Sia/pull/2191


tlightsky:
image

- add piece index
- host in one piece would be sorted and one line

PR #694, opened 2017-08-10
fix Receive Siacoin UX problems

### LINKED ISSUE #628: Change "Receive Siacoin" button to display the most recently generated address
I'm not sure if the UI devs follow the [/r/siacoin subreddit](https://www.reddit.com/r/siacoin/) actively, but probably the most common source of confusion for new users is the "Receive Siacoin" button in Sia-UI. Users not familiar with cryptocurrencies don't understand why the address is changing each time and some of them panic because they sent SC to the address shown and now they're seeing a totally different address.

Here are a few examples from just the last three weeks:

* [Why is the sia wallet address different each time?](https://www.reddit.com/r/siacoin/comments/6c5u3r/why_is_the_sia_wallet_address_different_each_time/)
* [Wallet address?](https://www.reddit.com/r/siacoin/comments/6etn6l/wallet_address/?st=j3qhq1qy&sh=b22842e1)
* [Does the wallet address change while syncing for the first time?](https://www.reddit.com/r/siacoin/comments/6c0fzm/does_the_wallet_address_change_while_syncing_for/?st=j3qhqeok&sh=a6184c0b)
* [[Help] Sent SC to the wrong address](https://www.reddit.com/r/siacoin/comments/6gb6dd/help_sent_sc_to_the_wrong_address/?st=j3qhqivz&sh=253e148b)
* [Sia-UI address constantly changing?](https://www.reddit.com/r/siacoin/comments/6eaoe1/siaui_address_constantly_changing/?st=j3qhqlu8&sh=a3df71d3)
* [Question about how SiaCoins Wallet works](https://www.reddit.com/r/siacoin/comments/6ewm1r/question_about_how_siacoins_wallet_works/?st=j3qhz6w5&sh=7a78e27a)
* ["Receive Siacoin" Addresses](https://www.reddit.com/r/siacoin/comments/6le1c4/receive_siacoin_addresses/)

Proposal: "Receive Siacoin" displays the most recently generated address. The user can still generate new addresses by going to the Terminal, but it eliminates a major source of confusion for new users.

Long term, I think there are ways to convey to the users that they have many addresses and can generate new ones on demand, but I hope this proposal is a reasonable short term solution.

avahowell:
this PR fixes the UX problems with 'Receive Siacoin', namely, that each time the dialog would open a new address would be displayed. This was confusing to new users unfamiliar with how sia works. The behavior has been changed so the dialog always shows the last generated address, and includes a button to generate a new address if desired.

Closes #628


Fornax96:
Not sure if this is going to get merged, since it goes against the privacy principles of the wallet. The "new address on every click" feature is intentional to prevent address re-use (https://en.bitcoin.it/wiki/Address_reuse). A better solution would be to have a separate screen with addresses and labels, and a receive screen where the user gets a new address and a text field to label it.

That way the user can see from who/what he received money when a transaction comes in. And it encourages generating a new address for every purpose. 

This is pretty much what every bitcoin / altcoin wallet does. Electrum also does it for example:
image
The description you enter in the receive form is used in both the address list, coin requests and in transaction log. This is how it should be.

I know that siad doesn't support labelling addresses itself, but this can trivially be implemented by using a little database (or even json file, who cares) in the Sia-UI directory, along with the config file et cetera.

lukechampine:
@Fornax96: We decided that this was a good compromise to address #628. Privacy is important to us, but it's not one of the primary goals of Sia, so we're ok sacrificing a bit of privacy-by-default for better UX. Users who want full privacy can still generate a new address for each transaction. Also, the backend will automatically generate new addresses for its own internal use (e.g. when forming new file contracts).

That said, I'm willing to consider moving towards the approach used by other wallets. We can address that in a separate issue/PR.

avahowell:
new behavior:
sia1
sia2

zachherbert:
Could this impact the experience of handling Siafund claims in the UI? I believe right now a user generates an address and then sends their SF to the new address. What would happen if a user tried to send SF to the same address, because they failed to click on the "New" button? Just thinking through edge cases.

avahowell:
I don't think Siafund claims should be affected by address reuse. cc @DavidVorick @lukechampine

plugins/Wallet/js/components/receiveprompt.js
```diff
@@ -1,19 +1,59 @@
...
+					<button className="new-address-button" onClick={handleGenerateClick}>New</button>
+				</div>
+				<h3> Prior Addresses </h3>
+				{ addresses.length > 0 ? (
+					<table className="pure-table address-table">
+						<tr>
+							<th>Description</th>
+							<th>Address</th>
+						</tr>
+						{ addresses.slice(0).reverse().map((oldAddress, key) => (
```
lukechampine: makes more sense for `addresses` to be a `List` so that it can be `reverse`d directly

PR #703, opened 2017-08-21
better allowance spending visualization

avahowell:
This PR improves Sia-UI's `UsageStats` component, updating it to use the renter's `financialmetrics` endpoint, and to visualize the various components of the allowance's spending. The simple `Spent / Allocated` is replaced with a segmented bar, where each spending metric (`upload`, `download`, `contract`, `storage`) is represented with a unique color and sized proportionately to its percentage of spending.

Screenshots:

sia1
sia2

### DIFF
--- a/plugins/Files/js/components/usagestats.js
+++ b/plugins/Files/js/components/usagestats.js
@@ -1,15 +1,38 @@
 import PropTypes from 'prop-types'
 import React from 'react'
 
-const UsageStats = ({allowance, spending}) => (
-	<div className="files-usage-info">
-		<span> {spending} SC Spent / {allowance} SC Allocated </span>
-	</div>
-)
+// UsageStats defines the presentation component for displaying file spending.
+const UsageStats = ({allowance, downloadspending, uploadspending, storagespending, contractspending}) => {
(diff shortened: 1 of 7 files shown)


avahowell:
Updated this PR so that the info panel informs the user when their allowance will renew.

plugins/Files/js/components/usagestats.js
```diff
@@ -1,15 +1,40 @@
...
 import React from 'react'
 
-const UsageStats = ({allowance, spending}) => (
-	<div className="files-usage-info">
-		<span> {spending} SC Spent / {allowance} SC Allocated </span>
-	</div>
-)
+// UsageStats defines the presentation component for displaying file spending.
+const UsageStats = ({allowance, downloadspending, uploadspending, storagespending, contractspending, renewheight}) => {
+	const unspent = () => allowance - (downloadspending+uploadspending+storagespending+contractspending)
```
lukechampine: why is this a function? Did it used to take an argument?

avahowell: I think so, not needed anymore

PR #724, opened 2017-10-06
rev 1 ui facelift: update side/topbar to new design

avahowell:
This PR updates Sia-UI's side and topbar to the new design. All of the plugins are left untouched. Still waiting for input from our designer re: fonts.

Screenshot:
newdesign


avahowell:
Latest screenshot:

sia

davecgough:
Have we considered a connections/bandwidth text/tool tip around synchronizing?

I know upon my first few uses I would close and reopen the program multiple times if I felt like it may have become "stuck".

PR #725, opened 2017-10-10
don't render usage stats component if allowance is empty

avahowell:


plugins/Files/js/components/usagestats.js
```diff
@@ -4,6 +4,9 @@ import React from 'react'
 // UsageStats defines the presentation component for displaying file spending.
 const UsageStats = ({allowance, downloadspending, uploadspending, storagespending, contractspending, renewheight}) => {
 	const unspent = allowance - (downloadspending+uploadspending+storagespending+contractspending)
+	if (allowance === 0) {
+		return <div />
+	}
```
lukechampine: don't we usually `return null` here?

avahowell: ah yeah

Issue #727, opened 2017-10-22
Negative funds remaining


nepeat:
# Bug
When the sum of the spending totals are more than the allowance, the UI glitches and displays negative SC remaining.

# Client/Server Info
Client is on macOS 10.13.0 GM, commit NebulousLabs/Sia-UI@d3aca71
Server is on Ubuntu Server 16.04 LTS, commit NebulousLabs/Sia@f16b5eb

# Screenshots
![App UI of the glitch](https://user-images.githubusercontent.com/6914822/31857897-ae1ddec6-b69e-11e7-84e1-6cf5ec028cc3.png)
![Zoomed crop for mobile -4.639999999999986 SC remaining](https://user-images.githubusercontent.com/6914822/31857909-13dc5bc0-b69f-11e7-972d-209eaf762fbf.png)

lukechampine:
Thanks for the detailed report!
What's the desired behavior here? Seems like the amount spent shouldn't exceed the allowance in the first place, but I guess we shouldn't assume that will never happen. Maybe display `4.63 SC over allowance` in red?

nepeat:
I believe there is a bigger issue with the UI display in the first place since the info that it is giving is very different than what the CLI gives me. 

I also was able to decrease my future allowance to 300 SC and well, the UI broke hard and couldn't display the allowance bar without rescaling the window.

Edit: Since GitHub sent my comment a little early, mind if I changed the bug title to "Allowance display problems when spending exceeds allowance"?

```
Renter info:
	Storage Spending:  118.3 SC
	Upload Spending:   3.697 SC
	Download Spending: 0 H
	Unspent Funds:     478.1 SC
	Total Allocated:   600.2 SC
```

### Overflowing into menu icons
![image](https://user-images.githubusercontent.com/6914822/31873754-217f9cd4-b77a-11e7-8b73-53130de20150.png)

### Contracts text cut off
![image](https://user-images.githubusercontent.com/6914822/31873758-2729013e-b77a-11e7-898d-0cb22cdbd485.png)

### Looks like it works, minus the SC remaining text
![image](https://user-images.githubusercontent.com/6914822/31873762-2c96ff72-b77a-11e7-9e7b-ba58c38e57d0.png)

avahowell:
[...]

PR #728, opened 2017-10-23
fix usagestats bar sizing, unspent calculation

avahowell:
See #727. This PR fixes an edge case where the spending bars could overflow their container, and changes Sia-UI to use the API's `unspent` instead of doing its own `unspent` calculation.


lukechampine:
LGTM but Travis is failing

avahowell:
Should be all set now.

PR #732, opened 2017-11-06
calculate and display download speed

huetsch:
I added a handy speed calculator for downloads, as shown in the attached screenshot. The code should be easily extensible to uploads, but I stopped with the download code because I found the "uploadprogress" field from the /renter/files API to be much less intuitive than the bytes "received" field from the /renter/downloads. Is there a reason why /renter/files has a percentage field instead of an "uploaded" analogue to "received"?

speed

### DIFF
--- a/plugins/Files/js/sagas/helpers.js
+++ b/plugins/Files/js/sagas/helpers.js
@@ -186,12 +186,61 @@ export const parseDownloads = (downloads) =>
 			})(),
 			siapath: download.siapath,
 			name: Path.basename(download.siapath),
+			bytestransferred: download.received,
+			totalbytes: download.filesize,
 			progress: Math.floor((download.received / download.filesize) * 100),
 			destination: download.destination,
 			type: 'download',
 			starttime: download.starttime,
 		}))
 		.sortBy((download) => -download.starttime)
+
+export const buildTransferTimes = (previousTransferTimes, transfers) => {
+	// Need a finite lookback window, lest we leak memory. This is also the window over which
+	// we calculate our average speed. This number can be adjusted.
+	const lookbackCount = 5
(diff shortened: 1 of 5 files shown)


lukechampine:
Very nice! This code looks pretty good to me, but I'll let @avahowell review it more thoroughly. Once the implementation is approved, we'll also want to add some tests.
Re: upload progress: it's a bit more complicated than downloads, because when you tell siad to upload a 1KB file, a lot more than 1KB is uploaded on the backend (depending on the redundancy settings). I agree that this is an area where the API could be better; hopefully we can address it in tandem with the overhaul of the renter that we're planning.

plugins/Files/js/sagas/helpers.js
```diff
@@ -186,13 +186,62 @@ export const parseDownloads = (downloads) =>
...
+			})
+		}
+		return map.set(transfer.siapath, {
+			timestamps: [Date.now()],
+			bytes: [transfer.bytestransferred],
+		})
+	}, Map())
+}
+
+const calculateSpeed = (transferTime) => {
```
avahowell: This should have a docstring, along with the previous func (`buildTransferTimes`), and the next (`addTransferSpeeds`). We're probably not completely consistent about this currently, but going forward I would like all UI helpers to have useful docstrings.

avahowell:
This code looks great, just needs two things:

- docstrings that help readers understand the algorithm used to compute download speed in the helper functions (`addTransferSpeeds`, `buildTransferTimes`, etc)
- test cases for the helper functions. Check out `test/files/helpers.js` for some examples of testing pure helpers.

Thanks for contributing!

huetsch:
OK, added docstrings and test cases.

PR #749, opened 2017-12-24
checkbox style bug

### LINKED ISSUE #747: "Hide OSC Transactions" checkbox active while unlocking wallet
The checkbox is active when the "Wallet Locked" popup is on top (tested on Windows). If I make the window small enough the checkbox is actually on top of the popup:
image

visgotti:


lukechampine:
This is being merged into the wrong branch -- should target `master`, not `wallet-css`

Issue #752, opened 2017-12-26
Possible open file descriptor leak


mckelvin:
## Environment
- OS: macOS
- Sia-UI/siad: reproduced in v1.3.1 and v1.3.0

## Problem

The `siad` inside the Sia-UI crashes daily without any error message. What's worse, other macOS apps will also crash. When I start the Sia-UI, it's OK in the beginning. But after one day, the siad crashes although the Sia-UI is still alive. At the same time, I can see a lot of crashed apps, which are not related to the Sia. I can't get out of the issue unless I quit the Sia-UI, otherwise other apps will keep failling.

<img width="736" alt="screenshot 2017-12-26 22 41 46" src="https://user-images.githubusercontent.com/865677/34358835-fab7c170-ea8d-11e7-8a9d-f8b841739c3c.png">

When the problem happens, there's no `siad` daemon and there're only some Electron processes(Only Sia-UI is using them). So I believe the bug is not in `siad`. From the logs listed below, I guess fd leaking are happening somewhere. I'm able to reproduce the issue 100% but each time it will take nearly 1 day to reproduce. Please let me know if you need more diagnosing info.

```
$ sudo lsof -n | cut -f1 -d' ' | uniq -c | sort | tail

 168 UserEvent
 211 Mail
 221 Dash
 242 AppleSpel
 264 socketfil
 266 Finder
 312 Google
 396 Dropbox
7958 Electron
```

```
$ lsof -d "^txt"
...
...
Electron  74202           kelvin   21r      REG                1,4     228354 121689443 /Applications/Sia-UI.app/Contents/Resources/electron.asar
Electron  74202           kelvin   22r      REG                1,4      48477  84153245 /Users/kelvin/Library/Application Support/Sia-UI/sia/consensus/consensus.log
Electron  74202           kelvin   23r      REG                1,4    1147674  84153242 /Users/kelvin/Library/Application Support/Sia-UI/sia/gateway/gateway.log
Electron  74202           kelvin   24r      CHR                3,2        0t0       304 /dev/null
Electron  74202           kelvin   25r      REG                1,4      48477  84153245 /Users/kelvin/Library/Application Support/Sia-UI/sia/consensus/consensus.log
Electron  74202           kelvin   26r      REG                1,4    1147674  84153242 /Users/kelvin/Library/Application Support/Sia-UI/sia/gateway/gateway.log
[...]

PR #754, opened 2017-12-27
Fix fd leak in readLog (#752)

mckelvin:
SEE #752 for details.


mckelvin:
As descripted in #752 ,this patch cannot resolve the issue. Close for now.

PR #791, opened 2018-06-18
Update packages through npm audit

eddiewang:
This PR updates some packages that could potentially be a vulnerability in Sia-UI.

It seems the most severe bugs are test related packages, with the exception of updating the electron binary.

When I built the app to test - I noticed that the hue was slightly off:
image

Upon further research, it appears that Chrome has some color rendering issues:
https://github.com/Microsoft/vscode/issues/51791

This was fixed by adding `app.commandLine.appendSwitch('force-color-profile', 'srgb');`

Here are the two screens side by side (right picture is before adding that line, and left being after adding the force-color-profilt):

image

The problem is that it still isn't accurate when compared to the current version. Here are all three screens side by side:

image

I'll continue investigating and decide on a direction to move forward with.


lukechampine:
not the first time our logo color has given us problems, and likely won't be the last. Oh well :/

PR #794, opened 2018-06-27
Initial UI Overhaul

eddiewang:
*ready for merge*
*ready for review*
This PR achieves the following:
 
Build System:
- Update packages that are out of date.
- Rework the webpack build pipline.
- Add webpack build pipelines for dev mode vs. prod.
- Use the standard JS lint with prettier-standard for formatting.
- Add git hooks with husky for lint checks before git commits.
- Update all js files with new lint rules.
- Make the html files automatically require between Dev and Prod js builds based on env.
- Hot reload for react components.
- CSS live reload for all plugins as well as the global css files.

UI:
- Rework Header and Sidebar to match new branding guidelines
- All core plugins updated with slightly more spacing for header components
- Core plugins updated to match new branding guidelines

The majority of the backend changes made are upgrades to the DX, which will hopefully make contributing to Sia-UI a nicer experience.

With this PR, developers can now run `npm run dev` to have a fully functional, hot reloadable Sia-UI app. Also, if contributors have different stylistic differences in the way they write code, the git hooks will automatically reformat them to standardjs rules on commit with husky.

Screenshots of the UI:
image
image
image


package.json
```diff
@@ -6,6 +6,38 @@
...
+  },
+  "lint-staged": {
+    "linters": {
+      "js/**/*.js": [
+        "prettier-standard",
+        "git add"
+      ],
+      "plugins/**/*.js": [
+        "prettier-standard",
+        "git add"
```
lukechampine: seems weird to me that `git` would be called directly from package.json. I'd rather have this command just do the linting, and let the user call `git add -u` separately.

lukechampine:
why did the package.json explode in size?

PR #795, opened 2018-06-27
Prettier Standard formatting

eddiewang:
This should be merged before #794 so I can rebase that branch with master. Thanks to @MSevey for the idea.

This PR is the simply the result of running `prettier-standard 'js/**/*.js' 'plugins/**/*.js'` in the repo.

Standard JS (https://github.com/standard/standard) is a popular style guide. I've used quite a few (Airbnb, Github, Google) and I always find myself coming back to Standard.

Prettier (https://github.com/prettier/prettier) is an auto-formatting tool that keeps code style consistent.


js/rendererjs/disabledplugin.js
```diff
@@ -28,10 +28,15 @@ const handleReport = () => {
...
+const DisabledPlugin = ({ errorMsg, startSiad }) => (
 	<div style={containerStyle}>
-		<h2>Siad has exited unexpectedly. Please submit a bug report including the error log <a style={reportStyle} onClick={handleReport}>here.</a></h2>
+		<h2>
+			Siad has exited unexpectedly. Please submit a bug report including the
+			error log{' '}
+			<a style={reportStyle} onClick={handleReport}>
+				here.
+			</a>
+		</h2>
```
lukechampine: hmm. Does this affect the actual display of this element? I feel like a JS formatter shouldn't be touching HTML.

plugins/Files/js/components/deletedialog.js
```diff
@@ -13,7 +13,8 @@ const DeleteDialog = ({files, actions}) => {
 			<div className="delete-dialog">
 				<h3> Confirm Deletion </h3>
 				<div className="delete-text">
-					Are you sure you want to delete {files.size} {files.size === 1 ? ' file' : ' files'}
+					Are you sure you want to delete {files.size}{' '}
+					{files.size === 1 ? ' file' : ' files'}
```
lukechampine: the `{' '}` looks wrong to me, given that the following strings are also prefixed by a space.

plugins/Files/js/components/directoryinfobar.js
```diff
@@ -20,13 +25,20 @@ const DirectoryInfoBar = ({path, nfiles, onBackClick, setDragFolderTarget}) => {
...
 				<i className="fa fa-backward" style={backButtonStyle} />
 				<span>Back</span>
 			</div>
 			<div className="directory-info">
-				<span style={{marginRight: '10px'}}> {path} </span>
-				<span style={{marginRight: '10px'}}> {nfiles} {nfiles === 1 ? 'file' : 'files' }</span>
+				<span style={{ marginRight: '10px' }}> {path} </span>
+				<span style={{ marginRight: '10px' }}>
+					{' '}
+					{nfiles} {nfiles === 1 ? 'file' : 'files'}
```
lukechampine: ditto

plugins/Files/js/components/unlockwarning.js
```diff
@@ -1,11 +1,16 @@
 import PropTypes from 'prop-types'
 import React from 'react'
 
-const UnlockWarning = ({onClick}) => (
+const UnlockWarning = ({ onClick }) => (
 	<div className="allowance-dialog unlock-warning">
-		<h1 className="unlock-warning-head"> Your wallet must be unlocked and synchronized to buy storage. </h1>
+		<h1 className="unlock-warning-head">
+			{' '}
+			Your wallet must be unlocked and synchronized to buy storage.{' '}
```
lukechampine: ditto

plugins/Files/js/components/uploaddialog.js
```diff
@@ -18,7 +18,10 @@ const UploadDialog = ({source, path, actions}) => {
 		<div className="modal">
 			<div className="upload-dialog">
 				<h1> Confirm Upload </h1>
-				<div>Would you like to upload {source.length} {source.length === 1 ? 'item' : 'items'}?</div>
+				<div>
+					Would you like to upload {source.length}{' '}
+					{source.length === 1 ? 'item' : 'items'}?
```
lukechampine: ditto

plugins/Files/js/components/usagestats.js
```diff
@@ -2,30 +2,70 @@ import PropTypes from 'prop-types'
...
+						{' '}
+						Storage Spending: {storagespending} SC{' '}
+					</li>
+					<li className="upload-spending-breakdown">
+						{' '}
+						Upload Spending: {uploadspending} SC{' '}
+					</li>
+					<li className="download-spending-breakdown">
+						{' '}
+						Download Spending: {downloadspending} SC{' '}
```
lukechampine: these too. I guess there's gonna be a lot. Bummer.

js/rendererjs/disabledplugin.js
eddiewang: note that this is jsx, not html. although it looks similar, there are some differences to how it works. It wouldn't affect the actual display of the element because jsx elements transpiles into javascript objects.

lukechampine:
this mostly looks fine to me. It's a bit aggressive on splitting up long lines (which traditionally we don't care about so much), and I also dislike indenting `case` blocks in `switch` statements; is the formatter configurable? I imagine we'd want to add it to our package.json somehow for easy invocation?

plugins/Files/js/components/deletedialog.js
eddiewang: {' '} represents a space. It's somewhat verbose but generally clearer to read because we're writing a mixture of JS and HTML in these files.

eddiewang:
The formatter is configurable - the scripts exist in `package.json` under `npm run format` as well as some git hooks on the larger PR, but for this one I just manually ran the script. I'm currently working on some failing tests that still exist on master. Once that's done, I'd like to rebase this PR so that Travis builds work properly again, so hold off on merging this for now.

plugins/Files/js/components/deletedialog.js
lukechampine: Sure, but in a lot of these cases we don't really care about the space. I think we often did stuff like:
```html
<h1> Some text here </h1>
```
to separate the content from the tags, but the formatter thinks that we want to render that extra space -- we don't.

eddiewang:
@lukechampine I removed the unnecessary whitespacing. You are right. Lots of whitespace that was accidentally being left in the HTML. But that's also why I prefer using {' '} for whitespacing in JSX - it makes it really clear that you have spaces in text.

The default width for long lines in Standard JS is 80 chars. I'm inclined to stick with the defaults (look at how clean the eslintrc is compared to the old one) for now, just because it's a well-known standard that people are used to. It's super easy to change that in the future, we just add a custom rule and run `npm run format`.

 I also updated the babelrc so that webpack builds properly, and so that tests passed based on my fix from yesterday's PR.

I've rebased from master and this PR should be good to go.

DavidVorick:
Is it able to distinguish between text lines and code lines?

Not a big deal / blocking factor for me either way. I would probably target 80 chars for text lines (since readability of a broken up line is not impacted) and 160 chars for code lines (since readability takes a huge hit if you split up a code line - unless you do it very elegantly, which I don't imagine a script is able to do.

PR #802, opened 2018-07-13
add appEntry check for renderer

eddiewang:
@lukechampine the fix was right, we just forgot to add a capital "A" to App.
 🤦‍♂️


lukechampine:
classic javascript

